PR #8, opened 2023-07-06
Merges conv1d and conv2d to Convolution

Jegp:
We presently represent dimensionality of the nodes implicitly via their weights. Except in the case of Convolution. 
This PR attempts to merge this, such that PyTorch `Conv1d` is represented as
```python
nir.Convolution(
  weights=np.zeros((1, 2, 3)),
  bias=np.zeros((3,)),
  ...
)
```

`Conv2d` would simply be expanding the weights and biases one dimension

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,36 +3,95 @@
 # If you wish to continue using _config.yml, make edits to that file and
 # re-generate this one.
 ###############################################################################
-author = 'Steven Abreu, Felix Bauer, Jason Eshraghian, Matthias Jobst, Gregor Lenz, Jens Egholm Pedersen, Sadique Sheik'
-bibtex_bibfiles = ['references.bib']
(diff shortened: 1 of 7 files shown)


sheiksadique:
Theoretically this is true but practically speaking pretty much every library has this implemented as separate modules, may be for some sensible reason? By merging these, we are essentially just adding more conditionality into our translation code for no real benefit except for the hypothetical case that some one will come up with a Conv4D.. 

This can only lead to higher code complexity and a higher number of bugs IMHO.

Jegp:
I think you have a point. I suggest closing this for now and then addressing it later, if needed. 

Ideally, I actually think the two worlds can coexist. In the sense that we can literally equate a `Convolution` and a `Conv1d` if the weights in the `Convolution` is 3D. Similar for 4D. How to achieve that in Python is a different matter...

PR #15, opened 2023-07-10
separated linear and affine layers

matjobst:
According to our discussions, I now separated the Linear and Affine layers to be:
Linear: y=Wx
Affine: y=Wx+b


stevenabreu7 (approved):
reformatted docstrings to make sure math is displayed correctly - otherwise all good for me!

PR #18, opened 2023-07-12
Add flatten node + add shape to output node

stevenabreu7:
still need to add new tests for flatten

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -143,10 +143,11 @@ class Delay(NIRNode):
 class Flatten(NIRNode):
     """Flatten node.
 
-    This node flattens the input tensor to a 1D tensor.
+    This node flattens its input tensor.
     """
 
-    n_dims: int = 1  # Number of dimensions to flatten
+    start_dim: int = 1  # First dimension to flatten
+    end_dim: int = -1  # Last dimension to flatten
 
 
 @dataclass
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -70,6 +70,14 @@ class Conv2d(NIRNode):
     groups: int  # Groups
     bias: np.ndarray  # Bias C_out
 
+    def __post_init__(self):
+        if isinstance(self.stride, int):
+            self.stride = (self.stride, self.stride)
+        if isinstance(self.padding, int):
+            self.padding = (self.padding, self.padding)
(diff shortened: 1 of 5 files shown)


nir/ir.py
```diff
@@ -120,6 +128,16 @@ class Delay(NIRNode):
     delay: np.ndarray  # Delay
 
 
+@dataclass
+class Flatten(NIRNode):
+    """Flatten node.
+
+    This node flattens the input tensor to a 1D tensor.
```
Jegp, line 135: Do we *always* want to flatten to a 1d tensor? Shouldn't this be specified somewhat like the PyTorch `Flatten` where you can select the dimensions to flatten?

For instance, something like `Flatten(start_dim, end_dim)`.

nir/ir.py
```diff
@@ -237,7 +255,7 @@ class Output(NIRNode):
     Defines an output of the graph.
     """
 
-    pass
+    shape: int  # Size of output
```
Jegp, line 258: I believe the shape should be an `np.ndarray`

Jegp (changes requested):
Neat additions! I have a few comments regarding the arguments of Input and the type of the `Output` shape parameter.

nir/ir.py
matjobst: For me the current version seems not really consistent or it needs more documentation. In the docstring, it says that the Flatten should flatten to a 1D Tensor. But then there is a parameter n_dims, "Number of dimensions to flatten". Is this supposed to select which dimensions should be flattened? Then the docstring is wrong. Or I am misunderstanding the parameter.
But I also like the proposal from @Jegp .

matjobst:
Maybe I missed the discussion. By my understanding the shape could be extracted from the node before the output node. For all other nodes (except Input, where it is necessary), we do not give the output shape. So why explicitly set the shape for the Output node?

biphasic:
@matjobst has a good point, I also don't see the point of having to specify dimension of the output node

Jegp:
> 

It'd be for convenience. Basically, a few people argued that it'd be helpful to directly read the shape directly from the graph instead of inferring it.

stevenabreu7:
very true, just updated to follow pytorch convention

PR #19, opened 2023-07-12
Named nodes

Jegp:
Replaces indices with string names

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -4,7 +4,7 @@ import typing
 import numpy as np
 
 
-Edges = typing.NewType("Edges", typing.List[typing.Tuple[int, int]])
+Edges = typing.NewType("Edges", typing.List[typing.Tuple[str, str]])
 
 
 class NIRNode:
@@ -21,7 +21,7 @@ class NIRGraph(NIRNode):
 
     A graph of computational nodes and identity edges."""
 
-    nodes: typing.List[NIRNode]  # List of computational nodes
+    nodes: typing.Dict[str, NIRNode]  # List of computational nodes
     edges: Edges
 
 
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -24,6 +24,14 @@ class NIRGraph(NIRNode):
     nodes: typing.Dict[str, NIRNode]  # List of computational nodes
     edges: Edges
 
+    @staticmethod
+    def from_list(*nodes: NIRNode) -> "NIRGraph":
+        """Create a sequential graph from a list of nodes by labelling them after indices."""
+        return NIRGraph(
+            nodes={str(i): n for i, n in enumerate(nodes)},
+            edges=[(str(i), str(i + 1)) for i in range(len(nodes) - 1)],
+        )
+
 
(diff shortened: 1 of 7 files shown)


nir/ir.py
```diff
@@ -100,7 +109,7 @@ class CubaLIF(NIRNode):
     r: np.ndarray  # Resistance
     v_leak: np.ndarray  # Leak voltage
     v_threshold: np.ndarray  # Firing threshold
-    w_in: np.ndarray = 1.  # Input current weight
+    w_in: np.ndarray = 1.0  # Input current weight
```
matjobst, line 112: This default value should be an np array.

tests/test_readwrite.py
```diff
@@ -27,121 +27,101 @@ def factory_test_graph(ir: nir.NIRGraph):
 
 
 def test_simple():
-    ir = nir.NIRGraph(nodes=[nir.Affine(weight=[1, 2, 3], bias=4)], edges=[(0, 0)])
+    ir = nir.NIRGraph(
+        nodes={"a": nir.Affine(weight=[1, 2, 3], bias=4)}, edges=[("a", "b")]
```
matjobst, line 31: There is no node "b", so this test should be failing!

nir/ir.py
```diff
@@ -21,9 +21,17 @@ class NIRGraph(NIRNode):
 
     A graph of computational nodes and identity edges."""
 
-    nodes: typing.List[NIRNode]  # List of computational nodes
+    nodes: typing.Dict[str, NIRNode]  # List of computational nodes
     edges: Edges
 
+    @staticmethod
+    def from_list(*nodes: NIRNode) -> "NIRGraph":
+        """Create a sequential graph from a list of nodes by labelling them after indices."""
```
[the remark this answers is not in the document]
Jegp: That's a good point! Would it be ok to defer that for now?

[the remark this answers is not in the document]
matjobst: Sure, I will create an issue!

tests/test_readwrite.py
Jegp: Hehe, true!

tests/test_ir.py
```diff
@@ -1,93 +1,100 @@
...
+    assert ir.nodes["in"].shape == [
         3,
     ]
-    assert ir.nodes[1].delay == [1, 2, 3]
-    assert ir.edges == [(0, 1), (1, 2)]
+    assert ir.nodes["d"].delay == [1, 2, 3]
+    assert ir.edges == [("in", "d"), ("d", "out")]
+
+
+def test_cuba_lif():
```
matjobst, line 72: This test does not test anything.
Additionally, the Ruff complains about the variable name 'l' which might be ambiguous with 1.

Jegp: Good catch. Sorry for the incomplete test.

nir/ir.py
Jegp: Addressed in #21

biphasic:
LGTM

Issue #21, opened 2023-07-12
Make all tests use numpy arrays


matjobst:
In the nodes, all parameters are defined as numpy arrays, however we currently do not use numpy arrays as parameters in the tests.

I propose we should update the tests to also use numpy arrays.

PR #23, opened 2023-07-12
Feature numpy tests

Jegp:
Addresses #21

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -250,11 +250,12 @@ class Output(NIRNode):
 
     pass
 
+
 @dataclass
 class Scale(NIRNode):
     r"""Scales a signal by some values.
 
-    This node is equivalent to the 
+    This node is equivalent to the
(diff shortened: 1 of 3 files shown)


Jegp:
This is rebased on #19 and should not be merged before, btw

matjobst:
The numpy changes are still missing in the read/write tests.

biphasic:
> The numpy changes are still missing in the read/write tests.

where in [here](https://github.com/neuromorphs/NIR/blob/main/tests/test_readwrite.py) are there still torch tensors used?

matjobst:
> > The numpy changes are still missing in the read/write tests.
> 
> where in [here](https://github.com/neuromorphs/NIR/blob/main/tests/test_readwrite.py) are there still torch tensors used?

I am not talking about torch tensors. But in NIR we basically said that we should use np.arrays for everything. Jens has already updated the tests in the test_nir to reflect that, but in test_readwrite there are still some python lists or scalar values used.

Jegp:
Nicely caught @matjobst . And thanks for the review @biphasic!

Issue #29, opened 2023-07-19
Add .inputs and .outputs properties to NIRGraph


matjobst:
For parsing NIR Graphs (especially when allowing multiple inputs and outputs), it might be useful to add .inputs and .outputs properties to the NIRGraph, such as follows:

```python
@dataclass
class NIRGraph(NIRNode):
    """Neural Intermediate Representation (NIR) Graph containing a number
    of nodes and edges.

    A graph of computational nodes and identity edges."""

    nodes: typing.Dict[str, NIRNode]  # List of computational nodes
    edges: Edges
    
    @property
    def inputs(self):
        return {name:node for name,node in self.nodes.items() if isinstance(node, Input)}
    
    @property
    def outputs(self):
        return {name:node for name,node in self.nodes.items() if isinstance(node, Output)}
```

If we have this example graph definition:
```python
 inner_model = nir.NIRGraph(
    nodes={
        "in1": nir.Input(np.array([4, 5, 2])),
        "in2": nir.Input(np.array([4, 5, 2])),
        "flat": nir.Flatten(0),
        "out1": nir.Output(np.array([20, 2])),
        "out2": nir.Output(np.array([20, 2])),
    },
    edges=[
        ("in1", "flat"),
        ("in2", "flat"),
        ("flat", "out1"),
        ("flat", "out2")
    ]
)
outer_graph = nir.NIRGraph(
    nodes = {
        "in": nir.Input(np.array([4,5,2])),
        "inner": inner_graph,
        "out": nir.Output(np.array([20,2]))
    },
    edges=[
        ("in", "inner.in1"),
        ("in", "inner.in2"),
        ("inner.out1", "out"),
        ("inner.out2", "out")
    ]
)
```

This allows retrieval of inputs or outputs, for example for the upper graph definition:
```python
inp0 = outer_graph.nodes['inner'].inputs['in0']
```

Or to iterate over all inputs/outputs:
```python
for name, outp_node in outer_graph.nodes['inner'].outputs.items():
    pass
```

stevenabreu7:
closed with #30

PR #30, opened 2023-07-19
implemented .inputs and .outputs properties for NIRGraphs

matjobst:
Implementation of Issue #29  "Add .inputs and .outputs properties to NIRGraph"

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -48,6 +48,18 @@ class NIRGraph(NIRNode):
     nodes: Nodes  # List of computational nodes
     edges: Edges  # List of edges between nodes
 
+    @property
+    def inputs(self):
+        return {
+            name: node for name, node in self.nodes.items() if isinstance(node, Input)
+        }
+
+    @property
+    def outputs(self):
+        return {
+            name: node for name, node in self.nodes.items() if isinstance(node, Output)
+        }
+
     @staticmethod
     def from_list(*nodes: NIRNode) -> "NIRGraph":
         """Create a sequential graph from a list of nodes by labelling them after
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -40,7 +40,7 @@ class NIRNode:
         return self is other
 
 
-@dataclass
+@dataclass(eq=False)
 class NIRGraph(NIRNode):
     """Neural Intermediate Representation (NIR) Graph containing a number of nodes and
     edges.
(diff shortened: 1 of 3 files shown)


nir/ir.py
```diff
@@ -23,6 +23,14 @@ class NIRGraph(NIRNode):
 
     nodes: typing.Dict[str, NIRNode]  # List of computational nodes
     edges: Edges
+    
+    @property
+    def inputs(self):
```
Jegp, line 28: I really appreciate this solution. It's really avoids any redundancy.

nir/ir.py
```diff
@@ -23,6 +23,14 @@ class NIRGraph(NIRNode):
...
+        return {name:node for name,node in self.nodes.items() if isinstance(node, Input)}
```
Jegp, line 29: I get a warning from ruff that the line is too long. May I suggest either adding pre-commit [as mentioned in the docs](https://nnir.readthedocs.io/en/latest/dev.html) or maybe bumping the minimum line length to something more than 88?

Jegp: Or maybe it's just a matter of rebasing onto main?

Jegp (approved):
Really nice work @matjobst. I particularly like how we're not adding any data to the existing declarations, but just using the existing data to provide convenient functionality.

There are a few linter errors, but that may just be a matter of rebasing to main?

Otherwise, LGTM!

matjobst:
I pushed this failing version now to showcase one issue that came up:
In the tests (test_ir.py), line 155 I check the following:
`assert ir.nodes["in2"] in ir2.nodes["inner"].inputs.values()`
However, that fails with "ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()"

The reason for that is that dataclasses generate a default comparison `__eq__()` function which compares all attributes of the class, but does not understand how to properly compare the arrays. For the case of the Input node, the resulting `__eq__()` does something like this:
```python
def __eq__(self, other):
    return self.shape == other.shape
```

Since the .shape is typically an array, it would rather need to do something like:
```python
def __eq__(self, other):
    return (self.shape == other.shape).all()
```

However I am still not sure whether we want that behavior. Because that means that:
```python
a = nir.Input(shape=np.array([1]))
b = nir.Input(shape=np.array([1]))
a == b   # True
```

This could be avoided by having a name in each node.

Jegp:
#50 stabilizes how input and output types are treated, so now would be a good time to merge this, I believe. @matjobst, would you have time to fast-forward this branch so we can proceed?

matjobst:
This is also a proposed fix to #36 
The equality of all nodes is replaced by identity.

Jegp (approved):
I think this is a solid implementation and fulfills the specs. Nice work!

Please go ahead and squash + merge when ready :)

Issue #26, opened 2023-07-14
Upload example code to `example/`


Jegp:
This issue tracks example code for each of the frameworks

- [ ] SpiNNaker2
- [ ] snnTorch
- [ ] Norse
- [ ] Sinabs
- [x] Nengo

PR #32, opened 2023-07-20
Example/sinabs

sheiksadique:
Added a quick example in the docs for sinabs - nir interaction

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -1,0 +1,5 @@
+pre-commit
+ruff
+black
+hdf5
+numpy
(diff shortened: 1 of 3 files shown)


biphasic:
I've already done that here https://github.com/neuromorphs/NIR/tree/main/example/sinabs
I think such examples should only exist once and then be linked if necessary

biphasic:
see also https://github.com/neuromorphs/NIR/issues/26

sheiksadique:
@biphasic Yes I have seen this example and pretty much copy pasted it here. But this is not referenced in the documentation esp. in the structure that @Jegp has now added. What do you suggest we do for consistency?

biphasic:
easiest way is to push a symlink. But actually if we integrate the examples in the documentation, we might be able to delete the top level examples

Issue #24, opened 2023-07-12
Support nir.__version__


matjobst:
Currently, the `nir.__version__` does not seem to work. We should support that.

biphasic:
for Sinabs and Tonic I'm using [pbr](https://docs.openstack.org/pbr/3.1.1/index.html) to automatically manage versions based on git tags. An option that I read about but have never tried is versioneer

sheiksadique:
I suggest we use [setuptools-git-versioning](https://setuptools-git-versioning.readthedocs.io/en/v1.13.3/) tool for version management if we want to automate and do intermediate/pre releases. https://setuptools-git-versioning.readthedocs.io/en/v1.13.3/

Jegp:
The latest PR #27 should add this to the `nir` package.

It uses the `dynamic` version number attribute in `pyproject.toml`, so the *only* place we have a version number is in the `__init__.py` file in the `nir` folder. 

I'm wondering if we should continue the discussion on release management. It would be nice to automate the release process. But then again, currently it's a simple matter of (a) bumping the version number and (b) creating a release. Is it worth improving? @biphasic @sheiksadique

sheiksadique:
I am not certain how you typically deal with release management, so I can't comment about its complexity.

But there are essentially two places where the version is maintained still: 
1. the __init__.py file
2. git tag/version.

It is therefore possible in principle to have these two things out of sync.

There are essentially three advantages to automate the process of releases. 

1. There will be a unique location of version information.
2. The releases themselves can be automatically handled ie you do not need to create a separate release. A simple act of tagging should automate a major/minor release.
3. Any versions that are not explicitly tagged but pushed to the repos can still be published as pre-releases giving users immediate access to this without having to look up things in GitHub.

Jegp:
Thanks for the comments @sheiksadique! I think you have very valid points, that would be worthy to treat in a separate issue: #48

PR #34, opened 2023-07-22
Added version to IR nodes and fixed linting

Jegp:
This PR addresses #24 by adding a `.version` field to every single `NIRNode`. This is also added to the file reading and writing.

Note that the version numbers have default values, so downstream APIs should _not_ need to change.

There is one slight wrinkle to this; Python dataclasses constructs an `__init__` function based on the (inherited) fields. And if a field with a default value is followed by a non-default field, the `__init__` method runs into trouble. This has been solved in Python 3.10 by adding a `kw_only` flag. But. For now, I'm using a hack from StackOverflow that creates a manual metaclass that then instantiates the Python dataclass. So, the outcome *should* be the same, but it's worth testing across multiple Python versions and OSs, I guess.


sheiksadique:
Adding a version field to every node seems like an overkill. Is it not sufficient to have this metadata at the top level of the saved file?

A version field for every node wrongly implies that you could have a file with nodes of multiple versions interspersed into one file.

Jegp:
> Adding a version field to every node seems like an overkill. Is it not sufficient to have this metadata at the top level of the saved file?

Yes, but how would we represent that? Given that the return type is a `NIRNode`, we would have to package that in there somehow.

What we could do instead is to tie the version number to the package. So, the only place we would represent it would be (a) in `__init__.py` and (b) in the files we write to/read from. The downside is that we would have to augment reading return to also specify a version number, since the reading can no longer just return a `NIRNode` (because that wouldn't contain the version number).

That leads me to consider adding a separate method (something like `read_version`) to check the version of the serialized object. The benefit to that would be that we (a) avoid storing the version number in the nodes themselves and (b) decouples the serialization of the graph from the reading of the version number. Would that approach address your concerns?

sheiksadique:
Ya I think it would be a good approach to add a separate metadata section
to the top level of the HDF file that consists of the versioning
information.

Ideally this shouldn’t break existing code.

Sadique

Jegp:
I removed the attribute as we discussed. Now the version is only readable from either `nir.version` or `nir.__version__`. Each generated hdf5 file consists of two root nodes: `node` and `version`.

Note that we can no longer tell the version number of a graph loaded into memory. But, that's likely not a problem because a graph loaded into memory relies on a very specific version of NIR. As far as I can tell, version conflicts will likely happen in two cases: when (old) files are read and when two libraries relies on different versions and exchange a NIR graph. The former can be identified by the written version number in the file. The latter can be resolved by dependency fixtures. I think it's reasonable to expect libraries to cling on to certain NIR versions to avoid breaking in the future.

Are you ok with the present model @sheiksadique?

sheiksadique:
Yes this sounds good to me. Thankyou!

Issue #28, opened 2023-07-19
Major Issue: Nodes with multiple inputs/outputs cannot be represented in the current format


sheiksadique:
The current graph representation assumes a single input/output for a node.

It doesn't have any mechanism to work or address specific inputs/outputs of a given module. This means the input/output nodes of a graph/subgraph have an unclear role/usage. This also renders subgraphs to be of very limited use.

There are two possible solutions to this:

1. Each node has a list of inputs and outputs. And for each output node have a list of edges that lead to specific input of a subsequent input node. Ie. the graph edge would look something like `NodeA.O[2] -> NodeB.I[0]`.
2. A second option would be to adopt ONNX/function like representation. `NodeB(NodeA.O[2], ... ) -> (NodeB.O[0], NodeB.O[1])`

Are there any proposals on how to deal with this?

matjobst:
I also like the subscript approach, but since we reference the nodes with their name when defining the edges, I would propose a fully string-based version that looks more object-oriented, so one edge would be for example ("in", "nodeB.in1"). That would signify an edge from the toplevel graph input "in" to the input "in1" of the subgraph "nodeB".

This is an example of how this could look like in practice: 
```python
 inner_model = nir.NIRGraph(
    nodes={
        "in1": nir.Input(np.array([4, 5, 2])),
        "in2": nir.Input(np.array([4, 5, 2])),
        "flat": nir.Flatten(0),
        "out1": nir.Output(np.array([20, 2])),
        "out2": nir.Output(np.array([20, 2])),
    },
    edges=[
        ("in1", "flat"),
        ("in2", "flat"),
        ("flat", "out1"),
        ("flat", "out2")
    ]
)
outer_graph = nir.NIRGraph(
    nodes = {
        "in": nir.Input(np.array([4,5,2])),
        "inner": inner_graph,
        "out": nir.Output(np.array([20,2]))
    },
    edges=[
        ("in", "inner.in1"),
        ("in", "inner.in2"),
        ("inner.out1", "out"),
        ("inner.out2", "out")
    ]
)
```

sheiksadique:
Thanks for the quick PR @matjobst . I have to highlight that this issue is not limited to Graphs but also Nodes because there could very well be nodes/modules that take in multiple inputs that have different meaning and then produce multiple values as outputs.

So the fix need to be extended to the Node definition perhaps and not just to the Graph

DylanMuir:
Hi all, I hope it's ok that I jump in!

[...]

PR #39, opened 2023-07-25
Dev/combinators

sheiksadique:
Added a node to address multiple/input/output of nodes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
       args: ["--profile", "black", "--filter-files"]
 
 - repo: https://github.com/myint/docformatter
-  rev: v1.6.0
+  rev: v1.7.5
   hooks:
     - id: docformatter
-      args: [--in-place, --wrap-summaries=88, --wrap-descriptions=88]
+      args: [--in-place, --black, --wrap-summaries=88, --wrap-descriptions=88]
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -303,24 +303,29 @@ class Threshold(NIRNode):
 
 
 @dataclass
-class Projection(NIRNode):
-    r"""Projects the input data to a different dimension.
+class Braid(NIRNode):
+    r"""Braids the input signals (wires) by providing a mapping from input indices to output indices.
+    The mapping does not have to be 1:1 (bijective): 
+    it can ignore signals (terminal objects), or duplicate signals, but it cannot create new signals
+    (initial objects).
+    Note that we assume the input signals are ordered. That is, an input signal will have a 
+    set of inputs that are numbered from 0 to N-1, where N is the total number of inputs.
 
     For instance:
 
-    Projection(output_indices=np.array([np.nan, 0, np.nan, np.nan, 1]))
+    >>> Braid(output_indices=np.array([np.nan, 0, np.nan, np.nan, 1]))
+    >>> (x, y) -> (0, x, 0, 0, y)
 
-    (x, y) -> (0, x, 0, 0, y)
+    If the given mapping contains a single number, it is interpreted to take the first value of *any*-size input.
+    That is, tuples of size one are considered individual values and not as tuples.
 
-    Tuples of size one are considered individual values and not as tuples.
+    >>> Braid(output_indices=np.array([0]))
+    >>> (x, y, z) -> x
 
-    Projection(output_indices=np.array([0]))
+    We can also duplicate the same signal twice:
 
-    (x, y, z) -> x
-
-    Projection(output_indices=np.array([np.nan, 0, np.nan, 0])
-
-    x -> (0, x, 0, x)
+    >>> Braid(output_indices=np.array([np.nan, 0, np.nan, 0])
+    >>> x -> (0, x, 0, x)
     """
 
     output_indices: np.ndarray
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -303,22 +303,22 @@ class Threshold(NIRNode):
 
 
 @dataclass
-class Project(NIRNode):
+class Projection(NIRNode):
     r"""Projects the input data to a different dimension.
 
     For instance:
 
-    Project(output_indices=np.array([np.nan, 0, np.nan, np.nan, 1]))
+    Projection(output_indices=np.array([np.nan, 0, np.nan, np.nan, 1]))
 
     (x, y) -> (0, x, 0, 0, y)
 
     Tuples of size one are considered individual values and not as tuples.
 
-    Project(output_indices=np.array([0]))
+    Projection(output_indices=np.array([0]))
 
     (x, y, z) -> x
 
-    Project(output_indices=np.array([np.nan, 0, np.nan, 0])
+    Projection(output_indices=np.array([np.nan, 0, np.nan, 0])
 
     x -> (0, x, 0, x)
     """
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -1,14 +1,14 @@
-from dataclasses import dataclass
 import typing
+from dataclasses import dataclass
 
 import numpy as np
-
 
 Edges = typing.NewType("Edges", typing.List[typing.Tuple[str, str]])
 
 
 class NIRNode:
     """Base superclass of Neural Intermediate Representation Unit (NIR).
+
     All NIR primitives inherit from this class, but NIRNodes should never be
     instantiated.
     """
@@ -16,18 +16,19 @@ class NIRNode:
 
 @dataclass
 class NIRGraph(NIRNode):
-    """Neural Intermediate Representation (NIR) Graph containing a number
-    of nodes and edges.
+    """Neural Intermediate Representation (NIR) Graph containing a number of nodes and
+    edges.
 
-    A graph of computational nodes and identity edges."""
+    A graph of computational nodes and identity edges.
+    """
 
     nodes: typing.Dict[str, NIRNode]  # List of computational nodes
     edges: Edges
 
     @staticmethod
     def from_list(*nodes: NIRNode) -> "NIRGraph":
-        """Create a sequential graph from a list of nodes by labelling them
-        after indices."""
+        """Create a sequential graph from a list of nodes by labelling them after
+        indices."""
         return NIRGraph(
             nodes={str(i): n for i, n in enumerate(nodes)},
             edges=[(str(i), str(i + 1)) for i in range(len(nodes) - 1)],
@@ -49,7 +50,7 @@ class Affine(NIRNode):
 
 @dataclass
 class Conv1d(NIRNode):
-    """Convolutional layer in 1d"""
+    """Convolutional layer in 1d."""
 
     weight: np.ndarray  # Weight C_out * C_in * X
     stride: int  # Stride
@@ -61,7 +62,7 @@ class Conv1d(NIRNode):
 
 @dataclass
 class Conv2d(NIRNode):
-    """Convolutional layer in 2d"""
+    """Convolutional layer in 2d."""
 
     weight: np.ndarray  # Weight C_out * C_in * X * Y
     stride: int  # Stride
@@ -152,7 +153,7 @@ class Flatten(NIRNode):
 
 @dataclass
 class I(NIRNode):  # noqa: E742
-    r"""Integrator
+    r"""Integrator.
 
     The integrator neuron model is defined by the following equation:
 
@@ -173,8 +174,8 @@ class IF(NIRNode):
         \dot{v} = R I
 
     .. math::
-        z = \begin{cases} 
-            1 & v > v_{thr} \\ 
+        z = \begin{cases}
+            1 & v > v_{thr} \\
             0 & else
         \end{cases}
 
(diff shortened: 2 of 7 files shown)


Jegp:
I think the name "Projection" is still a little misleading. I much prefer the mathematical / categorical nomenclature around "Braiding". Because that's precisely what it is! We're weaving threads together. And if there are any "loose ends", they are simply initial/terminal objects (0's in the current code). A final benefit is that it's much less ambiguous than "project/projection".

Would anyone be opposed if we used that term instead?

Jegp:
Following up on the discussion we had around dictionaries vs. tuples as types for inputs/outputs, let me quickly sketch the two scenarios using the following NIR graph:

```mermaid
graph TD;
A --> C;
B --> C
```

## 1: Dictionaries
Assume `C` has two ports for `"a"` and `"b"`. Then, it's input type dictionary would be something like:
`{"a": ..., "b": ...}`. But that means the edge coming *into* `C` would have to know which wire to connect where because dictionaries are not ordered (by default). So, there would have to be some form or "named braiding" that wires `A` to `"a"` and `B` to `"b"`. This would be similar to a function call, where we would very specifically set the inputs `f(a=..., b=...)`.

## 2: Tuples
Assume `C` has two ports. Then, the connection only needs to understand the arity because *any* two inputs, will work (assuming we don't type-check). So, the braiding would be "unnamed" in the sense that the wires feeding in would simply be a tuple `(..., ...)`. This has the disadvantage that we cannot name the inputs. Specifically, we can't say `f(a=..., b=...)`, we have to say `f(..., ...)`.

Curious to hear your thoughts.

sheiksadique:
Yes that is accurate. So once again, both work, we just have to settle on one that makes most sense. I believe the dictionary solution is more powerful/complete than Tuples.

stevenabreu7:
TL;DR
- dictionaries are nice - naming multiple inputs will avoid bugs and make NIR code more understandable
- most NIR graphs will be sequential, so making nodes with single I/O easy is important
- in code, braiding has more complicated syntax than named ports, while achieving the same
- code readability is important - NIR will be mostly used for generic network conversions

I don't know how dictionaries would be more powerful (given we'd only use flat dicts, so basically named tuples). But still, it's nice UX and will probably make NIR developers' lives easier. 

Here are some concrete NIR graphs that we will definitely have to support, first the simple case, then the more complicated one.

**1) Sequential architectures.** 
One input and one output for every node. I would imagine that the vast majority of NIR graphs in the near future will look like this.

```mermaid
graph TD;
Input --> Conv2d_a;
Conv2d_a --> ReLU_a;
ReLU_a --> AvgPool2d_a;
AvgPool2d_a --> AndSoOnAndOn;
AndSoOnAndOn --> Flatten;
Flatten --> Linear;
Linear --> ReLU;
ReLU --> Output;

Input_rnn --> Dense_in;
Dense_in --> LIF;
LIF --> Dense_rec;
Dense_rec --> LIF;
LIF --> Dense_out;
Dense_out --> Output_rnn;
```

**2) Example of a rSTDP-LIF population.**

```mermaid
graph TD;
Input_signal --> Dense_s;
Dense_s --> LIF_pre;
LIF_pre --> LearningDense;
LearningDense --> LIF_post;
LIF_post --> LearningDense;
Input_reward --> Dense_r;
Dense_r --> LearningDense;
```

_(we still need to add the Braids, i.e. all three inputs to LearningDense must go through their own Braid node)._

where now LearningDense has
- tuple signature `(array(N,), array(N,), array(N,)) -> (array(N,))`
- dictionary signature `{'input_spk': array(N,), 'reward_spk': array(N,), 'post_spk': array(N,)} -> {'output_spk': array(N,)}`. 

For the developer, this will look like 

```python
rstdp_demo = nir.NIRGraph(
    nodes={
        "input": nir.Input(
            shape=(array(N,)),
            shape={'...': array(N,)},
        ),
        "input_rew": nir.Input(
            shape=(array(N,)),
            shape={'...': array(N,)},
        ),
        "inp_dense": nir.Dense(
            input_shape=(array(N,)) or {"inp": array(N,)},
            output_shape=(array(N,)) or {"out": array(N,)}
        ),
        "rew_dense": nir.Dense(
            input_shape=(array(N,)) or {"inp": array(N,)},
            output_shape=(array(N,)) or {"out": array(N,)}
        ),
        "pre_lif:": nir.LIF(...),
        "pre_spk_braid": nir.Braid(
            output_indices=array([0, np.nan, np.nan])
        ),
        "post_spk_braid": nir.Braid(
            output_indices=array([np.nan, 0, np.nan])
        ),
        "rew_spk_braid": nir.Braid(
            output_indices=array([np.nan, np.nan, 0])
        ),
        "learning_dense": LearningDense(
            input_shape=(array(N,), array(N,), array(N,)),
            input_shape={
                'pre_spk': array(N,), 
                'post_spk': array(N,), 
                'rew_spk': array(N,)
            },
            output_shape=(array(N,)) or {"out": array(N,)}
        ), 
        "post_lif": nir.LIF(...),
        "output": nir.Output(...)
    },
    edges=[
        ("input", "inp_dense"),
        ("inp_dense", "pre_lif"),
        ("input_rew", "rew_dense"),
        ("rew_dense", "rew_spk_braid"),
        ("rew_spk_braid", "learning_dense"),
        ("pre_lif", "pre_spk_braid"),
        ("pre_spk_braid", "learning_dense"),
        ("learning_dense", "post_lif"),
        ("post_lif", "post_spk_braid"),
        ("post_spk_braid", "learning_dense"),
        ("post_lif", "output")
    ]
)
```

I think tuples or dictionaries don't make a big difference here. 

If we do go for dictionaries, it would be convenient to have some default key naming that we can use whenever we don't care about the key names (i.e., most of the time).

_(Something like `dd: tuple -> dict` with `dd(input_tuple) = dict(zip(def_keys[:len(input_tuple)], input_tuple))` such that `dd((input1, input2)) -> {'a': input1, 'b': input2}` (for some `def_keys='abcd...'`))_

To be fully exhaustive here, let's also take a look at this using named ports instead (similar to Matthias's proposal in #28, #29). This looks pretty much the same as above, but instead of using `nir.Braid`, we specify in the edges what ports to connect. By default, a NIR node can have one input and one output port, and then we don't have to specify any ports (neither in the node shapes, nor in the edges). We can assume that every I/O port accepts a single tensor.

```python
rstdp_demo = nir.NIRGraph(
    nodes={
        "input": nir.Input(shape=(N,)),
        "input_rew": nir.Input(shape=(N,)),
        "inp_dense": nir.Dense(
            input_shape=array(N,),
            output_shape=array(N,),
        ),
        "rew_dense": nir.Dense(
            input_shape=array(N,),
            output_shape=array(N,),
        ),
        "pre_lif:": nir.LIF(...),
        # can also define LearningDense as NIR graph, like in #28 
        "learning_dense": LearningDense(
            input_shape={
                'pre_spk': array(N,), 
                'post_spk': array(N,), 
                'rew_spk': array(N,)
            },
            output_shape=array(N,)
        ),
        "post_lif": nir.LIF(...),
        "output": nir.Output(...)
    },
    edges=[
        ("input", "inp_dense"),
        ("inp_dense", "pre_lif"),
        ("input_rew", "rew_dense"),
        ("pre_lif", "learning_dense.pre_spk"),
        ("rew_dense", "learning_dense.rew_spk"),
        ("learning_dense", "post_lif"),
        ("post_lif", "learning_dense.post_spk"),
        ("post_lif", "output")
    ]
)
```

To be honest, this looks a lot cleaner and simpler than the Braiding business while keeping all the functionality (right?), and we would spare developers from having to read and understand a "NIR Braiding tutorial". 

While it's true that only developers will ever use this (and not end users), it's also true that these devs will work with generic networks and won't be converting single concrete networks into NIR. This already makes debugging and testing a bit harder, so anything we can do to make the feedback/debugging loop faster and easier would save lots of pain.

Thoughts?

Jegp:
Excellent comment @stevenabreu7! 

>     * code readability is important - NIR will be mostly used for generic network conversions
Maybe we agree here, but let me quickly clarify that I actually don't believe that NIR *graphs* should be readable. The *code* should be readable (which I also think is what you're writing?) and it should be easy to use and create graphs (100%). Actually, I think that's one of the most important features. But. The graphs itself should foremost be *complete and unambiguous*. I think it's less essential that they are small than they are complete and disambiguates as much as possible.

With that opinion in mind, I'd like to turn to your point around braiding. I reckon that's essential to decide before we can take a decision on *how* to implement either braiding or ports.

>     * in code, braiding has more complicated syntax than named ports, while achieving the same
I agree that they essentially achieve the same thing, but I think the methods are different in one significant way. Ports use edges to couple nodes together. Meaning, the responsibility for sending signals to the correct places lies in the edges, not in the nodes. Conversely, braids uses nodes to rearrange the wiring internally. Meaning, the responsibility for sending signals to the correct places lies in the nodes.

Let me try to raise two issues around (a) compartmentalization and (b) disambiguation.

## Compartmentalization
Consider the following graph:
```mermaid
graph TD;
subgraph F
  C --> D;
end
subgraph G
  B --> F;
end
A --> G;
```
With ports, we can describe the connectivity as `[(A, G), (B, F), (C, D)]` because we know that `B` is the input node of `G` and `C` is the input node of `F`. But we'd also be able to subgraph into `G` with `(A, G.B)` and similarly with `(B, F.C)` because the `.` notation allows us to "look up" a node inside a subgraph. Essentially, `.` "exposes" the internal nodes of subgraphs, because the outside world needs to know that before they can (precisely) create edges to the subgraph. Now, what if that node itself if a subgraph? Then this *has* to be legal: `G.F.C`. 

I may be wrong, but this seems to follow from the necessity to "open up" subgraphs. And I'm not certain that it's a good idea to allow arbitrary lookups like this. Although, I'm not 100% convinced. Maybe it's fine to allow arbitrary nesting?

One thing I like about the braiding (but cannot formally prove why I like) is that it compartmentalizes that connections. If we, for instance, force connections to only be dictionary of named tensors, any node connecting to a subgraph doesn't have to know how the subgraph looks like, it just needs to know its "type signature". That is, which inputs does it have. (In some sense this is closer to the notion of "ports", oddly enough).

## Disambiguation of signals
Consider your graph on rSTDP-LIF. If we used ports, how would the LearningDense node distinguish from the inputs? We'd have edges looking something like `[(LIF_pre, LearningDense), (Dense_r, LearningDense), ...]`. In which order should they arrive? Or, which edge would map to which input in `LearningDense`? 

One solution would be to give names to the inputs of each subgraph, so we could say `(LIF_pre, LearningDense.a)` and so on. But then each edge would have to have information about the *type* of the target node, which sounds messy to me. Or, rather, that sounds like mixing up the responsibility of connectivity I mentioned above.

Curious to hear your thoughts.

stevenabreu7:
I see your point in the compartmentalisation with "opening up" subgraphs, I agree that this could become too messy. I was thinking we just allow the `.` notation only for `Input` nodes, so in that case, `G.B` works, but `G.F` doesn't work because `F` is not an `Input` node. (your last paragraph about braiding then just as well applies to named ports)

Re: the connectivity becoming a responsibility of the edges and not nodes. The edges will still only be "identity transformations", so I don't feel that they get more responsibility/compute. I think of edges with named ports as just "syntactic sugar" that is more concise than Braiding. If we use multiple `Braid` instances (as in the rSTDP example), the edges still need to know _which_ `Braid` to connect to, so won't the edges have just as much responsibility as in the named ports case? 

I don't think I understand the questions in "disambiguation of signals", are they still valid also for the "input-restricted `.` notation"?

Jegp:
Man. I'm sorry about the shitty writing. Let me try to clarify.

I see your point about "restricting" edges to subgraphs to only operate on input nodes. What worries me is that we now have "classes" of nodes. We would have to treat `Input` (and `Output`) as a very different thing. But maybe that's ok?

Regarding disambiguation, I'm trying to say that we would have to expose the types of the nodes *inside* the graph. That's not necessarily a problem, but I wonder whether that's a good separation of concern. I would disagree that using `Braid` puts responsibility on the edges because I cannot imagine a scenario where any node would have more than one braid. Meaning, `N` inputs will, at most need `N` braids that all output the same type. Essentially, the wiring is now happening inside the braid. To me, this is different from named ports because the wires wouldn't know any details about the source *or* destination of their connections. In ports, they would have to know the structure of the node, somehow.

This is getting hairy ;-) Wanna jump on a call to sort this out later today?

stevenabreu7:
in the example with the LearningDense (i.e., a non-nested node with complex input type), for making an edge we only need to know the `output_type` of the pre-node and the `input_type` of the post-node (same as for edges between nodes with simple input types)

but working with a nested graph, when defining the edges, we would need to also know what nodes are subgraphs, and what the name (and type) of its input nodes are. 

buuuuut, we also want to enforce that every NIRNode has an `input_type` and `output_type` defined, right? in that case, a NIR graph (being a NIR node) will also have to have an `input_type` and `output_type` (probably makes sense to compute that in the `__post_init__` function - creating an input_type which is a dictionary of `input_node_key -> input_node_shape`). 

so in the end, even with nested graphs, for making an edge, we still only need to know the `output_type` of the pre-node and the `input_type` of the post-node, right? so then the separation of concern is the same as with braiding?

Jegp:
> so in the end, even with nested graphs, for making an edge, we still only need to know the `output_type` of the pre-node and the `input_type` of the post-node, right? so then the separation of concern is the same as with braiding?

Ah, because you're assuming dictionaries as input types? Yes, in that case the input types of the subgraph would be sufficient to connect to, and they would correspond 1:1 to the input nodes in the subgraph.

I'd still argue that ports moves the concern of "connecting to the right input/from the right output port" into the edges. The braiding approach keeps this in the nodes.

Honestly, I can see the two approaches are doing mostly the same thing. I do think the braiding approach is slightly more "clean". More importantly, though, I think we should take a decision and move on :)

Jegp:
Closed in favour of #50

PR #40, opened 2023-07-25
fixed dev_requirements.txt

matjobst:
There was a typo in the `dev_requirements.txt`, requiring "hdf5" instead of "h5py".


Jegp (approved):
Nicely spotted

PR #41, opened 2023-07-26
edges as str when reading model from hdf5 file

matjobst:
When reading back a model from an hdf5 file, their endpoints will currently be read in as bytestring (`b""`), even though they were originally string (`str`).
This small change fixes it by enforcing them to be read back as `str`.


Jegp (approved):
This is all fine with me---it should make the code simple.

One thing that's slightly worrying to me is that none of our tests break from this change. Should we perhaps have a test that ensures that we can read nodes correctly, just to avoid some dumb regression down the line?

matjobst:
You are right, the read/write testcases did not check the edges at all. I modified the assert_equivalence checks to include checking the edges.

Issue #20, opened 2023-07-12
Generate meaningful names with from_list()


matjobst:
We could create more meaningful names instead of using the index as name:
For example similar to how Tensorflow handles that: It uses the type of node to create the names. For example:
The first instance of Dense will be called "dense", the second "dense_1" etc.

Jegp:
I see how it would make sense to have more meaningful names via the `from_list` constructor.

I'm not sure I'm a fan of the indexing, though. Doesn't that imply there is an ordering based on each node *type*? That `dense` comes before `dense_1`?

What about retaining the index across node types? Given the graph is constructed from a list, wouldn't it make sense to have a graph `Linear -> Linear` look like this? `Input_0` -> `Linear_1` -> `Linear_2` -> `Output_3`?

Concretely, it could look something like this:
```python
return NIRGraph(
    nodes={f"{type(n).__name__}_{str(i)}": n for i, n in enumerate(nodes)},
    edges=[(str(i), str(i + 1)) for i in range(len(nodes) - 1)],
)
```

matjobst:
Thank you for your proposal, that is of course another option. That has the nice benefit of not requiring a per-node-type counter.

However, it might also get confusing if you have a graph Dense->LIF->Dense. Then you would have `Input_0`->`Dense_1`->`LIF_2`->`Dense_3`->`Output_4`, with `Dense_1` and `Dense_3`, but no `Dense_2`.

In the Tensorflow-like naming, that would be `input`->`dense`->`lif`->`dense_1`->`output`.

I do not see the ordering implication as an issue here. But maybe that is also because I am used to the way Tensorflow does the default naming. Also this default naming is only for the from_list which can only create a sequential graph, giving a deterministic naming.

Addendum: In your proposal the edges also need the changed names :)
return NIRGraph(
    nodes={f"{type(n).__name__}_{str(i)}": n for i, n in enumerate(nodes)},
    edges=[(f"{type(nodes[i]).__name__}_{str(i)}", f"{type(nodes[i+1]).__name__}_{str(i)}") for i in range(len(nodes) - 1)],
)

Jegp:
I honestly don't have strong opinions here. I can see both systems working. If you're happy with the Tensorflow style, I would be fine to sign off on that :-)

PR #43, opened 2023-07-27
meaningful node naming when using from_list()

matjobst:
Implementation for Issue #20 
This should not break any end user code because it is just a modification on the naming of the nodes generated by NIRGraph.from_list().
Naming follows Tensorflow convention:
First occurence of Node type will get classname in lower case, from the second occurence a `_<i>` is appended, where `<i>` is the count of previous occurences.
Example:
input -> dense -> dense_1 -> affine -> dense_2 -> affine ->output

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -1,4 +1,5 @@
 import typing
+from collections import Counter
 from dataclasses import dataclass
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


nir/ir.py
```diff
@@ -30,9 +30,39 @@ class NIRGraph(NIRNode):
     def from_list(*nodes: NIRNode) -> "NIRGraph":
         """Create a sequential graph from a list of nodes by labelling them after
         indices."""
+
+        def unique_node_name(node, name_dict):
+            print(node)
```
Jegp, line 35: Remove print statements

nir/ir.py
```diff
@@ -30,9 +30,39 @@ class NIRGraph(NIRNode):
...
+            basename = node.__class__.__name__.lower()
+            print(basename)
```
Jegp, line 37: Remove

nir/ir.py
```diff
@@ -30,9 +30,39 @@ class NIRGraph(NIRNode):
...
+            idx = None
+            if basename in name_dict.keys():
```
Jegp, line 39: Another way to do this would be to have a defaultdict

```python
name_dict = defaultdict(int)
id = name_dict[basename]
name_dict[basename] += 1
node_name = f"{basename}" + f"_{id}" if id > 0 else ""
```

Or a [`Counter`](https://docs.python.org/3/library/collections.html#collections.Counter)
```python
appearances = Counter()
count = appearances[basename]
node_name = f"{basename}" + "_{count} if count > 0 else ""
appearances.update([basename])
```

Jegp (changes requested):
There are a few print statements that should be removed. And I had fun trying to find other ways to find a unique_node_name with some Python collection types.

Otherwise, nice addition  :)

matjobst:
Nicely spotted on the prints. They are removed now.
And thank you for the hint on the collections, I always forget about these. I implemented the Counter version now because it shows the intent of the code best.

Jegp (approved):
Nice! I like your way to do the format string with `f"{}{}"` it looks much better than `f"{}" + ...` :)

PR #46, opened 2023-08-11
Add BSD-3 LICENSE

stevenabreu7:
How does everyone feel about the BSD-3 license? Please give a quick thumbs-up if you agree. 

*BSD-3 is a permissive license comparable to the MIT license, but also adds that names of NIR contributors may not be used to endorse or promote products derived from this software without written consent - just an extra protection for NIR contributors. BSD-3 is used by [numpy](https://numpy.org/doc/stable/license.html), [sphinx](https://github.com/sphinx-doc/sphinx/blob/master/LICENSE), and [pytorch](https://github.com/pytorch/pytorch#license), so we would be in good company.*

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2023, NIR Team.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+   contributors may be used to endorse or promote products derived from
+   this software without specific prior written permission.
+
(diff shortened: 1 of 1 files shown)


jeshraghian:
👍

biphasic (approved):

sheiksadique (approved):

Jegp (approved):
I'm not a copyright lawyer, but BSD-3 seems like a good solution to me.

One concern would be related to patent law, where there are apparently differences between BSD3 and Apache v2 (https://fossa.com/blog/open-source-software-licenses-101-bsd-3-clause-license/). However, as far as I understand, patents are hard to issue for software, and since NIR is purely a software product, maybe that's not a concern?

Another brief comment is that "NIR Team" may be ambiguous. We do state it in the README, but would it make sense to spell out the original authors or link to our GitHub?

matjobst (approved):
Regarding the patent law comment: We previously had some discussions with a patent lawyer regarding software/algorithms. I hope I am not saying anything wrong here. In Europe you can only patent technical procedures/algorithms if they have some technical background, not software in general. And anyways a patent has to contain a certain amount of novelty and innovation. However, patenting anything already published is not possible. Not sure about the situation in the US or other places.
Because of these points, I anyways don't think patent law applies anymore.
The important thing is that we can not hold any responsibility for whatever anybody does with NIR and that anybody can use it for any purpose. For this BSD-3 seems like a good choice.

PR #49, opened 2023-08-21
Subgraph I/O with edges

Jegp:
This PR is another attempt to attack #28 that discusses how we can connect I/O from/to subgraphs.

Contrary to #39, this PR do not introduce a new `Project` construct, but instead uses edges to feed into/out of *parts* of the node. To make that explicit, I added two fields to *all* nodes: `shape_input` and `shape_output`. The fields describe exactly what the shape going in and out of the node will be. This allows any outside node to connect to a *part* of the node using indexing (`node[index]`). This generalizes to all nodes (including subgraphs), so if there are multiple inputs expected, we can index directly into that. 
As an example, consider the following graph
```mermaid
    graph TD;
    A --> C;
    B --> C;
```

I C has input shape `[2, 1]`, say, then an edge can directly contain information about where to inject the input, like so: `("A", "C[0]")`. For B: `("B", "C[1]")`.

The benefit of the approach is here is that we keep the information about the computation (node) separate from the connections (edges). Specifically, we leave the information about how signals connect out of the nodes, so the only place we can describe connections will be in the edges.

I drew inspiration from the [Rockpool](https://rockpool.ai/advanced/graph_overview.html) and [Lava-DL](https://lava-nc.org/) implementations.

Note that the implementation is a proof of concept and incomplete. Currently, the read/write tests are failing and I did not implement the `shape_input` and `shape_output` for all the nodes.

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -1,6 +1,6 @@
 import typing
 from collections import Counter
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 import numpy as np
 
@@ -14,6 +14,9 @@ class NIRNode:
     All NIR primitives inherit from this class, but NIRNodes should never be
     instantiated.
     """
+
+    shape_input: np.ndarray = field(init=False, kw_only=True)  # Shape of input
+    shape_output: np.ndarray = field(init=False, kw_only=True)  # Shape of output
 
 
 @dataclass
@@ -40,19 +43,22 @@ class NIRGraph(NIRNode):
             return name
 
         counts = Counter()
-        node_dict = {}
+        node_dict = [
+            ("input", Input(shape=nodes[0].shape_input)),
+        ]
         edges = []
 
         for node in nodes:
             name = unique_node_name(node, counts)
-            node_dict[name] = node
+            node_dict.append((name, node))
 
-        names = list(node_dict)
-        for i in range(len(nodes) - 1):
-            edges.append((names[i], names[i + 1]))
+        node_dict.append(("output", Output(shape=nodes[-1].shape_output)))
+
+        for i in range(len(node_dict) - 1):
+            edges.append((node_dict[i][0], node_dict[i + 1][0]))
 
         return NIRGraph(
-            nodes=node_dict,
+            nodes={n[0]: n[1] for n in node_dict},
             edges=edges,
         )
 
@@ -68,6 +74,13 @@ class Affine(NIRNode):
     """
     weight: np.ndarray  # Weight term
     bias: np.ndarray  # Bias term
+
+    def __post_init__(self):
+        assert len(self.weight.shape) >= 2, "Weight must be at least 2D"
+        self.shape_input = self.weight.shape[:-2] + tuple(
+            np.array(self.weight.shape[-2:]).T
+        )
+        self.shape_output = self.weight.shape[:-2] + (self.weight.shape[-2],)
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


matjobst:
How would this look like when you have multiple inputs of larger shapes? For example for element-wise binary operations (something like an adder y=a+b):
input a: shape=(32,64,64)
input b: shape=(32,64,64)

stevenabreu7:
> How would this look like when you have multiple inputs of larger shapes? For example for element-wise binary operations (something like an adder y=a+b):
> input a: shape=(32,64,64)
> input b: shape=(32,64,64)

I think then y would just have the edges `('a', 'y[0]')` and `('b', 'y[1]')`? And y would have `input_shape=(2,32,64,64)`, right?

matjobst:
That makes sense, I like that. I have two more:
Assume you want to add a scalar value to a tensor: y=a+b, where a.shape=(32,64,64), b.shape=(1,), using broadcast rules like in numpy. I mean this could still be done by adding lots of edges. But is there a more efficient way?
Or a graph that takes two inputs of completely different shapes, e.g. a.shape=(32,64,64), b.shape=(3,13,17,19). Things like this can happen when doing multi modal networks.
How could this look like?

stevenabreu7:
I generally like this approach. However, it assumes that the inputs are all of the same shape right? As @matjobst points out, there are some cases when we would like to have inputs of different shapes feed into the same node, so forcing the inputs to be of the same shape would be a limitation. 

To avoid this, we would need something a bit closer to what @matjobst suggested in #28 with string indexing. The connections could look something like `('F', 'C[filter]')` and `('A', 'C[data]')` where the implementation could be something like
```
class CrazyConv(NIRNode):
  inputs: Dict[str, np.ndarray]  # {input_name: input_shape}
  outputs: Dict[str, np.ndarray]
```
(this is somewhat similar to the InPorts/OutPorts in lava)

We could also stick with the assumption that inputs are of the same shape, but I'm afraid that would lead to some ugly work-arounds where this functionality is needed.

matjobst:
I would like to add one slight downside of the named I/O proposal: Concatenating and splitting tensors is not possible that way, so we would need to add dedicated nodes for that. Which I would be fine with.

stevenabreu7:
Yeah we'd need an explicit way to merge/split data. We can do this with nodes like in #39 or with edges (where the edges do some indexing like `(A, C[data][:10])`). 

If we go with #39, we keep our edges simple (which I think is a good idea). Plus, we can enforce that every input port is connected to only one output (if there are more connections, we would need to sum/concat the inputs - but we will enforce this to be done explicitly). A single output port can however be connected to many input ports (no aggregation needed).

This adds more nodes into our graphs, but down the line we could have some Nengo-like visualization that would also visualize these "routing" nodes nicely.

matjobst:
I also want to recall that we previously defined that when multiple edges go into one node input, the data should be added up (which is only clearly defined if they are of the same shape). So if we have _a->c_ and _b->c_ than the outputs of _a_ and _b_ should be added before being processed by _c_.

matjobst:
Having both name and index in array notation form could be ambiguous if they are optional. Assume data[0]. Does it access the name "0" or the index 0? Hence for this case I would propose using the "." notation for the name, such as:
(A, C.data[:10])
Then data.0 is referring to the name "0", where data[0] refers to the index 0.

Jegp:
Closed in favor of #39

PR #50, opened 2023-09-01
add input/output shape annotations for all nodes

stevenabreu7:
we want to have `input_shape` and `output_shape` properties on all NIR nodes. 

in many cases, these shapes can be easily computed from the node itself. if this is possible for a node, this is done in the `__post_init__` method.

if it's not possible to infer the input and output shape from the node directly, then the two fields `input_shape` and `output_shape` are required when creating the node. if these are not passed into the node's constructor, an error is thrown.

**design decisions to be made**
- do we want to enforce input shapes and output shapes? i.e. throw if they are not given? (currently only the case for `Flatten`)
  - alternative: we can allow uninferrable input/output shapes in a node, and then infer them when a NIR Graph is constructed (this allows us to infer the `Flatten` I/O shapes based on its pre- and post-nodes)
- how to implement `input_shape` and `output_shape` for every node? for now I went with option ii) - up for debate
  1. as optional in NIRNode → inherited by all nodes. fails bc they become the first two arguments for all nodes
  2. add input_shape and output_shape into every NIRNode subclass. _(implemented here)_
  3. as optional in NIRNode, and make all nodes `@dataclass(kw_only=True)`. works, but: we always have to initialize everything with keywords (no shortcut as in `ir.Threshold(0.5)`, must be `ir.Threshold(threshold=0.5)`.
  4. make the input_shape and output_shape @properties. this becomes tricky when for nodes where the shapes are not easily inferred from its given data (i.e., `Flatten`).

I'd love some feedback on these two questions. thanks!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
 
     runs-on: ${{ matrix.os }}
 
--- a/README.md
+++ b/README.md
@@ -23,9 +23,19 @@ Each computational unit is a node in a static graph.
 Given 3 nodes $A$ which is a LIF node, $B$ which is a Linear node and $C$ which is another LIF node, we can define edges in the graph such as:
 
 ```mermaid
-graph RL;
+graph LR;
 A --> B;
 B --> C;
+```
+
+Or more complicated graphs, such as
+
+```mermaid
+graph LR;
+A --> A;
+A --> B;
+B --> C;
+A --> C;
 ```
 
 ## Format
--- a/README.md
+++ b/README.md
@@ -15,15 +15,18 @@ The goal of NIR is to provide a common format that different neuromorphic framew
 ## Computational primitives
 > Read more about in our [documentation about NIR primitives](https://nnir.readthedocs.io/en/latest/primitives.html)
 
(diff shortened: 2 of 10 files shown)


Jegp:
Alright. I thought a bit about our setup, and I came to the realization that the "ports" and "braid" approaches are achieving the same thing. As far as I can see, the one difference is that the braiding node removes the responsibility of the edges to know which "port" to slot into. But, the braiding node will then very explicitly map the ports, one for each input or output. And, in the case of braids, each "coupling-braid" will have to know the type of the subgraph. Conversely, I fear that the "ports" approach requires (recursive) introspection into the node, so that the *edges* will have to know what's inside the subgraphs.

However, I believe we can find a compromise, where we still use the clean 1:1 mapping in an edge we always agreed upon, but where we insist that each node has an `input_type` and `output_type`. This is useful for subgraphs because we can immediately check for type compatibility with edges. In the example of a subgraph `G` we can directly type check the edge `A -> G.Input`, for instance. Easy!

I also believe that it's important to ensure that all node *types* are equivalent. And if the subgraph has a dict as its input/output type, then I think we should stick to *always* use dictionaries of named types (`Dict[str, np.ndarray]`) for input/output connections. This means that *every node has named arguments* in their inputs and outputs. The types are no longer one or more numpy arrays, they are now maps from strings to arrays in all nodes. Since NIR is an intermediate representation, I don't really mind, though, because it's easy to create syntactic sugar around the dicts to make them easy to work with. Currently, I've encoded input types to "regular" nodes as `{"input": ...}` and outputs as `{"output": ...}`. So, in the case where we have singular inputs/outputs we can just allow edges without subscripts `A -> B` to actually mean `A.output -> B.input`. What do you guys think?

Here are a few changes to the code from Steve
- I'm using model (i) where `input_type` and `output_type` is shared between *all* `NIRNode`s. I used the `field(kw_only=True)` to avoid the positional argument error. This means that it's possible for users to overwrite them if they'd like to. Importantly, it unifies *all* the nodes to force the input/output type fields.
- I enforced the `input_shape` argument for the `Flatten` in the spirit of disambiguation.
- I changed the input/output type of Linear/Affine to allow arbitrary-dimensional mappings. But I'm uncertain if it's correct. I'll leave it here for now and double-check for tomorrow, but I would appreciate a second pair of eyes :)

Looking forward to hearing your inputs!

stevenabreu7:
Exactly, this was the discussion we started in #39. 

- I pushed one change to Affine and Linear, I think it works now - but we should add some more tests for n-dimensional arrays.
- I also allowed naming the input_shape dict key for `Flatten` anything (right now it fails if it's not `input`). Alternatively, we can also add into the `Flatten` docstring that its dict key must be `input`, and then assert that.

Also, doing this `field(kw_only=True)` requires Python 3.10. Are we okay with enforcing that? Alternatively, we can stick with the (clumsier) way that I coded up (option (ii) in this PR's top comment).

stevenabreu7:
remaining tests fail because of the python 3.10 dependency

PR #51, opened 2023-09-05
Improved argument parsing for shapes

Jegp:
Added a helper function to allow simpler passing of shapes argument for input, flatten, and output nodes

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -1,5 +1,6 @@
 import typing
 from collections import Counter
+from collections.abc import Sequence
 from dataclasses import dataclass
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


nir/ir.py
```diff
@@ -11,6 +12,15 @@
 # Shape is a dict mapping strings to shapes
 Shape = typing.Dict[str, np.ndarray]
 
+def _parse_shape_argument(x: Shape, key: str):
+    if isinstance(x, np.ndarray):
+        return {key: x}
+    elif isinstance(x, Sequence):
+        return {key: np.array(x)}
+    elif isinstance(x, dict):
+        return x
```
stevenabreu7, line 21: perhaps here we can also assert that the dict has only one key which is the key passed into this function (i.e. "input")? in other places we assume that the only input key is "input", and this will break unless we enforce that here.

stevenabreu7 (commented):
otherwise looks good to me!

nir/ir.py
Jegp: That's a good point. I'll leave it for now, just to move forward here. In principle, I think these kinds of assertions are meaningful, but I'm cautious to add them because they can break people's code (maybe someone wants to violate that some day?) and they take a bit of runtime overhead. The last point is probably minimal. But still :)

Jegp:
Thanks for the review @stevenabreu7 ! :partying_face:

PR #52, opened 2023-09-07
Improved docs on types and renamed `input_shape` to `input_type`

Jegp:
This PR 
* Adds developer docs for nodes, edges, and node typing
* Renames `input_shape` and `output_shape` to `input_type` and `output_type` following yesterday's discussion

### DIFF
--- a/tests/test_architectures.py
+++ b/tests/test_architectures.py
@@ -21,6 +21,16 @@ def test_sequential():
 
 
 def test_two_independent_branches():
+    """
+    ```mermaid
+    graph TD;
+    A --> B;
+    B --> C;
+    C --> D;
+    E --> F;
+    F --> G
+    ```
+    """
     # Branch 1
     a = mock_affine(2, 3)
     b = nir.Delay(np.array([0.5, 0.1, 0.2]))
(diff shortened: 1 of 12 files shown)


tests/test_architectures.py
```diff
@@ -273,13 +319,13 @@ def test_complex():
...
         v_threshold=np.array([1, 1, 1]),
     )
-    # TODO: This should be a node that accepts two inputs
+    # TODO: This should be a node that accepts two input_type
     d = nir.Affine(
         weight=np.array([[[1, 3], [2, 3], [1, 4]], [[2, 3], [1, 2], [1, 4]]]),
         bias=np.array([0, 0]),
     )
     e = nir.Affine(weight=np.array([[1, 3], [2, 3], [1, 4]]), bias=np.array([0, 0]))
-    # TODO: This should be a node that accepts two inputs
```
stevenabreu7, line 282: should I implement this? (ie with a simple NIR graph that takes two inputs and puts them through different Affine nodes) - or what's the plan with it?

stevenabreu7 (approved):

tests/test_architectures.py
Jegp: I reckon the plan was to provide a node that actually uses two different inputs, instead of having two inputs that connect to the same "port".

I suggest we merge it for now, but leave the TODOs for later :)

Jegp:
Thanks for the review @stevenabreu7 ! :partying_face:

PR #54, opened 2023-09-14
53 ir for sumpool2d

sheiksadique:
Added implementation for SumPool2d

### DIFF
--- a/docs/source/primitives.md
+++ b/docs/source/primitives.md
@@ -15,7 +15,7 @@ $$
 \text{Linear} &:  \mathbb{R}^{m \times n} \\
 \text{Conv1d} &:  \mathbb{R}^{c_{out} \times c_{in} \times x},  \text{Strides}, \text{Groups}, ... \\
 \text{Conv2d} &:  \mathbb{R}^{c_{out} \times c_{in} \times y \times x},  \text{Strides}, \text{Groups}, ... \\
-\text{SumPool2d} &: \matchbb{R}^{c \times y \times x}, \text{Pooling}, \text{Strides}, \text{Padding} ...\\
+\text{SumPool2d} &: \mathbb{R}^{c \times y \times x}, \text{Pooling}, \text{Strides}, \text{Padding} ...\\
 \text{Threshold} &:  \begin{cases} 1 & v > v_{threshold} \\ 0 & else \end{cases}
 \end{align}
 $$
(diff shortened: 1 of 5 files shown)


docs/source/primitives.md
```diff
@@ -15,6 +15,7 @@ $$
 \text{Linear} &:  \mathbb{R}^{m \times n} \\
 \text{Conv1d} &:  \mathbb{R}^{c_{out} \times c_{in} \times x},  \text{Strides}, \text{Groups}, ... \\
 \text{Conv2d} &:  \mathbb{R}^{c_{out} \times c_{in} \times y \times x},  \text{Strides}, \text{Groups}, ... \\
+\text{SumPool2d} &: \matchbb{R}^{c \times y \times x}, \text{Pooling}, \text{Strides}, \text{Padding} ...\\
```
Jegp, line 18: Think this should be `\mathbb{R}`, no?

But this should be updated and formatted anyway with the table from the paper, I think.

nir/ir.py
```diff
@@ -166,6 +166,19 @@ def __post_init__(self):
...
+@dataclass
+class SumPool2d(NIRNode):
+    """Sum pooling layer in 2d."""
+
+    kernel_size: np.ndarray  # (Height, Width)
+    stride: np.ndarray  # (Height, width)
+    padding: np.ndarray  # (Height, width)
+
+    def __post_init__(self):
+        self.input_type = {"input": ()}
```
Jegp, line 178: Could we do something similar to https://github.com/neuromorphs/NIR/blob/main/nir/ir.py#L143 ?

Jegp (changes requested):
Is it reasonable to provide a proper inference of the input/output shapes? I know there will be unknowns, but I do believe we can be a bit more specific.

nir/ir.py
sheiksadique: The shape of the input cannot be inferred from the parameters of pooling or convolutional layers .. at least in the case of conv2d, we can specify the channels of the input and output. Height and width are arbitrary. (it is incorrectly defined in the post init code of conv2d btw.) In case of pooling, there is no constraint on any of the dimensions. The most you can say is that the input needs to be 2/3 dimensional.

This is also why I am not warn to the notion of adding these things in the node description.

Jegp: I see your point about the shapes. My thinking comes from the "disambiguate as much as possible"-principle. I agree with what you're saying, but I also don't have any good ways to resolve this, I find. This doesn't break anything, so maybe we should just get going? And then think about how to deal with shapes going forward.

Jegp (approved):
LGTM :money_with_wings:

PR #59, opened 2023-10-06
Specify spatial shapes in convolutions

Jegp:
As @matjobst pointed out, the convolutions incorrectly uses the weights to infer the spatial shape of the convolutions.
Second, they do not correctly serialize and deserialize.

This PR fixes the shapes of the convolutions to address both. However, it means that we are *forcing* the convolutions to be of a certain shape. I believe you, @sheiksadique, argued against this, so I would be curious to hear your opinion on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ If you use NIR in your work, please cite the [following Zenodo reference](https:
   month        = jul,
   year         = 2023,
   publisher    = {Zenodo},
-  version      = {0.0.1},
+  version      = {0.2},
   doi          = {10.5281/zenodo.8105042},
   url          = {https://doi.org/10.5281/zenodo.8105042}
 }
--- a/docs/source/about.md
+++ b/docs/source/about.md
@@ -37,7 +37,7 @@ If you use NIR in your work, please cite the [following Zenodo reference](https:
   month        = jul,
   year         = 2023,
   publisher    = {Zenodo},
-  version      = {0.0.1},
+  version      = {0.2},
   doi          = {10.5281/zenodo.8105042},
   url          = {https://doi.org/10.5281/zenodo.8105042}
 }
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -219,7 +219,8 @@ class NIRGraph(NIRNode):
         of nodes in the graph as needed. Assumes that the input_type of the graph is set. Moves
         from the input nodes to the output nodes. Raises ValueError if types are inconsistent.
 
-        Assumes that all input types are of form: {'input': ...} and all output types are of form: {'output': ...}.
+        Assumes that all input types are of form: {'input': ...} and all output types are of form:
+        {'output': ...}.
 
         Currently only supports the inference of output types for Conv1d and Conv2d nodes.
         Does not support nested NIR graphs.
@@ -247,7 +248,7 @@ class NIRGraph(NIRNode):
             ])
             if undef_post_input_type:
                 # define post input_type to be the same as pre output_type
-                print(f'[warning] {post_key}.input_type undefined -> set to {pre_key}.output_type')
+                print(f'[warning] {post_key}.input_type undefined, set to {pre_key}.output_type')
                 post_node.input_type = {
                     k.replace('output', 'input'): v for k, v in pre_node.output_type.items()
                 }
@@ -371,7 +372,7 @@ class Conv1d(NIRNode):
 
     def __post_init__(self):
         if isinstance(self.padding, str) and self.padding not in ["same", "valid"]:
-            raise ValueError(f"padding must be 'same', 'valid', or an integer, not {self.padding}")
+            raise ValueError(f"padding must be 'same', 'valid', or int, not {self.padding}")
         if self.input_shape is None:
             # leave input and output types undefined
             self.input_type = {"input": None}
@@ -427,7 +428,7 @@ class Conv2d(NIRNode):
 
     def __post_init__(self):
         if isinstance(self.padding, str) and self.padding not in ["same", "valid"]:
-            raise ValueError(f"padding must be 'same', 'valid', or an integer, not {self.padding}")
+            raise ValueError(f"padding must be 'same', 'valid', or int, not {self.padding}")
         if isinstance(self.padding, int):
             self.padding = (self.padding, self.padding)
         if isinstance(self.stride, int):
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -24,40 +24,56 @@ def _parse_shape_argument(x: Types, key: str):
 
 
 def _calculate_conv_output(
-    x_in: typing.Union[int, np.ndarray],
-    padding: typing.Union[int, np.ndarray],
-    dilation: typing.Union[int, np.ndarray],
-    kernel_size: typing.Union[int, np.ndarray],
-    stride: typing.Union[int, np.ndarray],
-) -> int:
+    input_shape: typing.Union[int, typing.Sequence[int]],
+    padding: typing.Union[int, str, typing.Sequence[int]],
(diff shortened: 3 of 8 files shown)


matjobst:
Checking further, that could also extend to SumPool and Flatten. Currently SumPool2d just has empty tuples as input and output shapes, Flatten has arrays with only zero entries (this is still without this PR, from the paper CNN):
input <class 'nir.ir.Input'> {'input': array([ 1,  2, 34, 34])} {'output': array([ 1,  2, 34, 34])}
0 <class 'nir.ir.Conv2d'> {'input': array([2, 5, 5])} {'output': array([16,  5,  5])}
1 <class 'nir.ir.IF'> {'input': array([ 1, 16, 16, 16])} {'output': array([ 1, 16, 16, 16])}
2 <class 'nir.ir.Conv2d'> {'input': array([16,  3,  3])} {'output': array([16,  3,  3])}
3 <class 'nir.ir.IF'> {'input': array([ 1, 16, 16, 16])} {'output': array([ 1, 16, 16, 16])}
4 <class 'nir.ir.SumPool2d'> {'input': ()} {'output': ()}
5 <class 'nir.ir.Conv2d'> {'input': array([16,  3,  3])} {'output': array([8, 3, 3])}
6 <class 'nir.ir.IF'> {'input': array([1, 8, 8, 8])} {'output': array([1, 8, 8, 8])}
8 <class 'nir.ir.Flatten'> {'input': array([0, 0, 0, 0])} {'output': array([0, 0, 0])}
9 <class 'nir.ir.Affine'> {'input': array([128])} {'output': array([256])}
10 <class 'nir.ir.IF'> {'input': array([  1, 256])} {'output': array([  1, 256])}
11 <class 'nir.ir.Affine'> {'input': array([256])} {'output': array([10])}
12 <class 'nir.ir.IF'> {'input': array([ 1, 10])} {'output': array([ 1, 10])}
output <class 'nir.ir.Output'> {'input': array([ 1, 10])} {'output': array([ 1, 10])}

Additionally, most layers include a batch dimension (IF, Input, Output) where others do not (Affine).

stevenabreu7:
I think batch sizes should generally not be included in our shapes since 1) batch sizes are variable and 2) most neuromorphic HW does not batch. 

I don't have a strong preference between specifying the node input/output types  or leaving them undefined for the ones in question (pool, conv, flatten). But I do think always specifying them is cleaner. We could also infer them when a NIRGraph is created (in `NIRGraph.__post_init__`).

sheiksadique:
Here is my take on this whole discussion about shapes and any other information that we include into the NIR.

1. First and foremost: NIR should have *sufficient* information to be able to replicate *function* on any given platform.
2. For the user to deploy this *function* with some *data*, the user needs to have some information about what *shape* it should have.
3. Information that is being *inferred* is not *necessary* and can only lead to redundant or incorrect information that could have been inferred else where if needed.

In conclusion, shapes are only useful within the context of NIR when they cannot be inferred but necessary.

A convolution operation does not require shape of its input. For instantiation of *neruons* and its parameters on hardware perhaps the shape is needed, but a spiking layer is different and has this information. A convolutional layer doesn't need to hold this information.

You can extend my comment to say that `__post_init__` is not necessary as the job of this function is *inferring* based on existing data.

sheiksadique:
With regards to batching I agree with you all that perhaps we shouldn't include it in the parameters (states) and their shapes should therefore correspond to single batch scenario.

stevenabreu7:
I agree, this is a nice way to frame it - as NIR only providing the information necessary for the represented computation. 

The inferred shapes are not necessary, I agree, but if we don't compute them in NIR then some platforms will end up writing code to infer the shapes - leading to redundancy. I think it would be nice for devs using NIR if we directly provide that code. 

I don't have a strong preference about whether we just provide the code for inferring shapes, or also run it by default. Right now, I don't see any practical application where _not providing_ inferred shapes would be advantageous, so I would just infer and provide them directly.

matjobst:
> A convolution operation does not require shape of its input. For instantiation of _neruons_ and its parameters on hardware perhaps the shape is needed, but a spiking layer is different and has this information. A convolutional layer doesn't need to hold this information.

If you completely follow this argumentation, you basically do not *need* to give the shape for anything. If the neuron has parameters of full shape, then you have all information that you need. If they don't, you could still infer based on the data which you input into the neuron.

But in my opinion, the function is tightly coupled to the type of data you could feed into the network and hence I see the shapes as integral part of the network (even though maybe redundant). A network that we export or import is made for a specific type of data. One advantage of explicit shapes is that it allows for checking for errors and simplifies the conversion. It also allows a kind of human readability (for any network I develop in TF, the model.summary() is on of the first things I do, to see how the network looks like and what the shapes are).

If we do not have the shapes in the NIR, the first step when importing from NIR to SpiNNaker2 would have to be the inference of all shapes throughout the network. I find this to be highly redundant when the information is already available on export to NIR.

At any point, if one wishes to use differently shaped input data with the same network (I do not even see this as a real use case as it would most likely require different parameters throughout the network), one is still free to ignore or modify the shape information.

sheiksadique:
I will leave this decision to made democratically and will settle for what ever is decided but I will try to make my argument one last time.

To put it all in context, when we all define pytorch or tensorflow models, we don't define the exact shape for every layer do we? This is because it is not necessary to do so and can be inferred. 

I agree with @matjobst that having shapes and a model summary is very helpful for development and I actually even have this information extracted in the NIRTorch library when extracting models. I am only suggesting here that this information can be retrieved with the inclusion of input shapes, without having to store intermediate shapes. I can sympathise with the fact that Spinnaker platform requires these shapes for each layer, but other platforms/libraries might not. I don't think it is a great idea to include information that each platform needs into NIR itself.

I will point out that NIR is not really "human readable" util the file is loaded using a software program. You can't just load a hdf file using a text editor. After loading with an appropriate software (python in our case), any number of things can be done to improve its interpretability or usability. So I agree that we should provide software tools to make this loading and interpretation easier and as useful as possible. 

I am arguing that adding information that isn't necessary in the NIR specification itself can possibly lead to inconsistencies (like the one @matjobst  caught about shapes of convolutional layers). The tools can continue to improve and provide as much information as we find necessary for downstream libraries but by introducing such errors in the NIR file itself will make it harder to fix (without regenerating the file itself).

bvogginger:
Note that currently the information of the input shape is required in the `nir.Input` node.
While not required -- in practice every NIR graph will have an Input node so that the shape information of subsequent layers such as Conv2D, SumPool2D etc. can be deduced.

One solution could be to offer a helper function that adds the shape information to all nodes in the NIR graph. This could be used in a `model_summary` offered by NIR.

matjobst:
@bvogginger I like this proposal. Then the information does not have to be directly embedded into the NIR, but is still available for those who need it. And not everyone needs to write their own shape propagation.

stevenabreu7:
I like this solution as well, I will code it up now

stevenabreu7:
Just pushed code to 
- allow padding to be a string of either 'same' or 'valid' (as supported in PyTorch, TF, etc)
- check that all type annotations are valid via `NIRGraph._check_types()`
- infers types within a NIR graph via `NIRGraph.infer_types()` (currently supports type inference for going forward, where the graph's input type is defined and then types are propagated forward through the graph. going backwards from the graph's output type can also still be done - if needed?)
- added tests for a simple ConvNet NIR graph with undefined types
Any feedback/suggestions are welcome!

Jegp:
Super neat changes. I like how this will be much easier for the user of the library.

There are probably some edge-cases we haven't thought about, but you did some thorough testing, so I suggest we just move forward and then see what can be improved. Even though we all know that @stevenabreu7 writes flawless code :angel:

PR #60, opened 2023-10-11
Fix type annotations for convolutional layers

stevenabreu7:
If this is okay then I will update all the other docstrings to be in the same style. 

Question: do we want to also allow for padding = 'same' or 'valid' in the Conv1d and Conv2d classes? These are the only values supported by Nengo atm.

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -145,7 +145,21 @@ class Affine(NIRNode):
 
 @dataclass(eq=False)
 class Conv1d(NIRNode):
-    """Convolutional layer in 1d."""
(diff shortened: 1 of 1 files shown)


stevenabreu7:
Just saw that @Jegp already did part of this in #59 - should we decide whether to support padding = 'same' | 'valid', and then implement that in #59?

matjobst:
Approve. For SpiNNaker2, 'same' and 'valid' are already supported :)

stevenabreu7:
closing this PR in favour of #59 where support for this is added

PR #61, opened 2023-10-12
enforce the validity of NIR graphs on creation

stevenabreu7:
PR #59 introduces code to check the consistency and "defined-ness" of input and output types across a NIR graph (also pointed out in #4). This is currently not enforced. 

However, NIR, as an intermediate representation, would do well to enforce the correctness of all created NIR graphs. Otherwise, we end up with potentially many invalid NIR graphs that will effectively be un-parseable. 

This PR enforces the validity of NIR graphs, as defined by the `NIRGraph._check_types()` method. If the graph types are undefined or inconsistent, a warning is printed and an attempt is made to infer them with `NIRGraph.infer_types()`. If this fails, an error is thrown. 

TODO:
- fix tests

_Note: this can also be merged only after our paper deadline._

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -176,6 +176,13 @@ class NIRGraph(NIRNode):
         self.output_type = {
             node_key: self.nodes[node_key].output_type for node_key in output_node_keys
         }
+        # check that all nodes have consistent and defined input and output types
+        try:
+            self._check_types()
+        except ValueError as e:
+            print(f'[warning] {e}')
+            self.infer_types()
+            self._check_types()
     
     def _check_types(self):
         """Check that all nodes in the graph have input and output types. Will raise ValueError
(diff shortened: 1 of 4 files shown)


Jegp (changes requested):
This sounds like a solid idea, but may I suggest adding a flag somewhere so the user can choose to ignore the pass? Maybe a parameter like `ignore_type_check = False`?

stevenabreu7:
Good idea, I've added a flag as NIRGraph parameter (`ensure_validity`). I've also put a `strict` flag into the read and write functions for NIRGraphs. This is turned on by default (since people should be using this), but can be turned off. 

Many tests are failing now, because we had a lot of inconsistent graph types defined in the tests..

Jegp:
I think this is similar to what #123 is doing; enforcing type checks on graph creation. I'll close this for now, but feel free to reopen if we ever need a `valid` flag in the graph somewhere.

PR #63, opened 2023-10-12
Add v_reset to neuron models

stevenabreu7:
For the LIF and CNN task, we currently use a fixed reset voltages - which is counter to the docstrings and parameterisation of the `LIF`, `IF`, and `CubaLIF` models (thanks to Terry for pointing that out). 

This PR adds a `v_reset` parameter to these neuron models. By default it is None, and the subtractive resetting described in the docstrings right now is applied. If `v_reset` is set to some value, then the voltage is set to this value when a spike is emitted.

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -476,6 +476,14 @@ class CubaLIF(NIRNode):
             v-v_{threshold} & z=1 \\
             v & else
         \end{cases}
+    
+    or, if the reset voltage is not None:
+
+    .. math::
+        v = \begin{cases}
+            v_{reset} & z=1 \\
+            v & else
+        \end{cases}
 
     Where :math:`\tau_{syn}` is the synaptic time constant,
     :math:`\tau_{mem}` is the membrane time constant,
@@ -484,6 +492,7 @@ class CubaLIF(NIRNode):
     :math:`v_{leak}` is the leak voltage,
     :math:`R` is the resistance,
     :math:`v_{threshold}` is the firing threshold,
+    :math:`v_{reset}` is the reset voltage,
     and :math:`S` is the input spike.
     """
 
@@ -493,6 +502,7 @@ class CubaLIF(NIRNode):
     v_leak: np.ndarray  # Leak voltage
     v_threshold: np.ndarray  # Firing threshold
     w_in: np.ndarray = 1.0  # Input current weight
+    v_reset: typing.Optional[np.ndarray] = None  # Reset voltage
 
     def __post_init__(self):
         assert (
@@ -502,6 +512,8 @@ class CubaLIF(NIRNode):
             == self.v_leak.shape
             == self.v_threshold.shape
         ), "All parameters must have the same shape"
+        if self.v_reset is not None:
+            assert self.r.shape == self.v_reset.shape, "All parameters must have the same shape"
         # If w_in is a scalar, make it an array of same shape as v_threshold
         self.w_in = np.ones_like(self.v_threshold) * self.w_in
         self.input_type = {"input": np.array(self.v_threshold.shape)}
@@ -594,15 +606,26 @@ class IF(NIRNode):
             v-v_{thr} & z=1 \\
             v & else
         \end{cases}
+    
+    or, if the reset voltage is not None:
+
+    .. math::
+        v = \begin{cases}
+            v_{reset} & z=1 \\
+            v & else
+        \end{cases}
     """
 
     r: np.ndarray  # Resistance
     v_threshold: np.ndarray  # Firing threshold
+    v_reset: typing.Optional[np.ndarray] = None  # Reset voltage
 
     def __post_init__(self):
         assert (
             self.r.shape == self.v_threshold.shape
         ), "All parameters must have the same shape"
+        if self.v_reset is not None:
+            assert self.r.shape == self.v_reset.shape, "All parameters must have the same shape"
         self.input_type = {"input": np.array(self.r.shape)}
         self.output_type = {"output": np.array(self.r.shape)}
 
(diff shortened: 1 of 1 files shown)


bvogginger:
Thanks @stevenabreu7 for providing this PR.

I wonder whether it would be better to *explicitly* specify the reset mechanism.
e.g. `reset = "subtract"` or `reset = "v_reset"`.

How different frameworks implement this:
- sinabs: provide reset function: Callable()
- snntorch: flag for reset mechanism
- spinnaker2: flag for reset mechanism
- spikingjelly: as in this PR.

Instead of a flag one could also introduce a bool:
```py
reset_is_subtract: bool = True
if spike:
    v = reset_is_subtract*(v-v_threshold) + (1-reset_is_subtract)*v_reset
```

Other opinions?

stevenabreu7:
Yeah, both are fine for me. As you point out, both exist in other tools and it'll be clear enough either way as long as the docs explain it well. 

If we use the flag, we would have to either 1) set a default value for `v_reset` (e.g. zero, but then we'd have to cast the zero into an array of the correct shape in the __post_init__ - a bit messy), or 2) keep the default value for `v_reset` as `None` (but then it's _possible_ to set the mode to "reset" and have `v_reset=None` which would have to throw an error - which seems fine to me for an IR). 

Any other opinions?

bvogginger:
Ok, considering the potential issues during initialization when using a *flag*, I tend to go for the existing solution provided by @stevenabreu7 !

Jegp:
Hmmmm. I think this is a hard issue. Formally in NIR (in the paper at least), the behavior of the reset is defined to be $v-v_{thr}$. What you're suggesting would change that, effectively change the underlying dynamics and parameterize the equations. Which concerns me because it goes against my intuition that our primitives are "primitive".

But, I also see the point of being able to describe the exact reset mechanism in NIR. Would it be meaningful to instead of adding it into the specific neuron models define that

1. We agree to use a standard reset mechanism in the neuron models by default (could be $v - v_{thr}$, but could also be with the reset in this PR)
2. But that we can compose neuron dynamics with different kinds of resets. If we, for instance, wanted a reset to zero, we could create a LIFZero primitive which would essentially be $\text{Reset}_{zero} \circ LI$.

Jegp:
> With 1), we would have to change existing frameworks to use a different reset mechanism or even add it if it doesn't exist yet (like in Norse for I&F). Are we okay with "setting/enforcing standards" like that?

In some sense, the value of NIR is exactly this kind of standardization. It will ruffle some feathers, but at some point we have no choice but to unify the descriptions of the computations. We should ensure our reasons our sound, and we pick the right representation, of course.

My main concern with the flag solution is that one primitive all of a sudden will represent multiple continuous time systems. Now it's no longer just a LIF. It's a LIF-reset or LIF-subtract.

> If we go for 2), I'm afraid we might end up with a bunch of different LIF variants which will clutter the number of primitives that we support, making it more confusing and unclear.. (and I think it wouldn't be as simple as combining a new `reset_zero` primitive with `LI`, because `reset_zero` would have to act on the membrane potential of `LI`, which is an internal variable and can neither be read nor set from the outside, at least not the way we set things up now).

That's true. But how is the parameterization different? Wouldn't any backend have to ensure they support all the different threshold mechanisms?
A separate point is that I'm not super clear on how many mechanisms we're talking about. I've heard of the $v = v - t_{th}$ (subtract) and $v = 0$ (reset). Are other variants in use?

Separately, it would be interesting to understand exactly how different the two mechanisms would be. Perhaps it'd be worthwhile running some experiments?

stevenabreu7:
I see your point that LIF should represent _one_ kind of system, and not multiple different ones like `LIF-reset` _and_ `LIF-subtract`. Having separate primitives will also make it easier for us to specify what platforms support which primitives. 

I agree that standardisation is nice, but I also don't feel confident enough in any model to restrict to NIR to only that one. I don't even think there is a "right" model or representation. I do think there are more different ways to implement the reset. For example, allowing graded spikes (multiple spikes per timestep) further complicates this because then the reset can be either `v-v_th` or `v-n*v_th` where `n` is the number of spikes emitted. 

We likely won't get a conclusive answer here and now. How should we move forward? Should I split the `LIF` into `LIF_reset` and `LIF_subtract` and create an alias `LIF = LIF_reset` for backward compatibility?

bvogginger:
Note that this issue of `"reset-to-v_reset"` vs. `"subtract"` only applies to time-step based simulation but not the the continuous-form we present in the paper.

In the continuous form, when the voltage reaches the threshold, it is immediately reset to 0 by subtracting the threshold (or to `v_reset` for this PR). In the continuous dynamics the voltage never exceeds the threshold. Hence **subtracting the threshold** and **resetting to 0** are **identical**!

If we want to stick to NIR being a continuous time representation of "neuromorphic networks", then there is only one unique description.
Note that I would prefer then to use the `v_reset` version as presented in this PR, as it is more general. E.g. it allows to reset the membrane voltage below `v_leak` as is the case for most of biological neurons. (Note that mathematically it is sufficient to specify either `v_leak` or `v_reset` and let the other one be 0.).

The problem only applies to time-step based simulations such as those in the ML-frameworks or digital neuromorphic hardware. It does not apply analog neuromorphic hardware.

Hence, one solution could be to let the user choose the reset method when running `from_nir(graph, reset_method)`, as this is an implementation detail.

We also could add metadata to the exported graph that specifies the reset mechanism used. Similarly, one could specify the used integration method, e.g., "Forward Euler" or "Exponential Euler".

I'm afraid that this leads to a more general discussion on whether NIR is a continuous dynamics format for Neuormorphic architectures, for for exchanging time-step based SNN between ML-Frameworks...

stevenabreu7:
that's absolutely correct, thanks for this comment @bvogginger. I opened a new issue to go forward with this, and I will close this PR.

PR #68, opened 2023-10-17
Documentation update

biphasic:
* Moved examples into ipython notebooks and rendered them as part of the documentation. They are statically rendered now (execution is set to 'off'), but once the library is stable, we can have them automatically run as part of the CI pipeline. That ensures that tutorial code actually works! Another possibility is then to combine our examples folder with something like Binder, which then has all those notebooks available to play with
* Copied over the primitives currently described on overleaf into a table format
* Used the nice NIR schema figure from overleaf in the 'what' section
* Used sphinx auto-api, which generates a much nicer API reference
* Light logo in the sidebar, some text fixes here and there

Difference between [before](https://nnir.readthedocs.io/en/latest/) and [after](https://nnir.readthedocs.io/en/docs-update/)


Jegp (approved):
Wow, @biphasic, this is a great contribution. I really liked how you cleaned up the docs and the setup for the docs. Wonderful, thank you so much!

Jegp:
Regarding the execution, I think you have a great point about directly providing *runnable* examples. I also have high hopes for stuff [like thebe](https://jupyterbook.org/en/stable/interactive/thebe.html) to run the notebooks directly in the browser. It's not quite there, but we should definitely keep an eye out for ways to make the code as accessible as possible.

PR #71, opened 2023-10-19
Bug fixes: nir.Output type inference & CubaLIF.w_in read/write

stevenabreu7:
~Right now, the type inference goes through but it doesn't actually change the NIR graph. This PR fixes that.~

Update: this PR fixes a weird edge case where the input_type of a `nir.Output` node is inferred. This currently fails because it will not update the output_type for the `nir.Output` node and this is used when writing the graph to a file. (hence, when writing to a file and reading again, the shape inference is essentially discarded). 

Update update: I've also added an urgent fix that adds `w_in` to the read and write method of the `CubaLIF` node. This should have always been there..

### DIFF
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -261,6 +261,12 @@ class NIRGraph(NIRNode):
                     k.replace('output', 'input'): v for k, v in pre_node.output_type.items()
                 }
 
+            # make sure that output nodes have output_type = input_type
+            if isinstance(post_node, Output):
+                post_node.output_type = {
+                    k.replace('input', 'output'): v for k, v in post_node.input_type.items()
+                }
+
             # check if post output_type needs to be defined
             undef_post_output_type = post_node.output_type is None or any(
                 v is None for v in post_node.output_type.values()
--- a/nir/ir.py
+++ b/nir/ir.py
@@ -285,6 +285,9 @@ class NIRGraph(NIRNode):
             seen.add(post_key)
             ready += [e for e in self.edges if e[0] == post_key and e[1] not in seen]
 
+            self.nodes[pre_key] = pre_node
+            self.nodes[post_key] = post_node
+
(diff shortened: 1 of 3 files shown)


matjobst:
For me the inference changed the graph already without your modifications. Did you observe the nodes not changing in a specific instance?
The dataclasses are mutable unless set to frozen. That is why it is not necessary to actively write the nodes back into the graph.

stevenabreu7:
yep I just wrote tests to check that this works, but the problem was something else. I will update the PR in a moment - sorry about the confusion

nir/ir.py
```diff
@@ -261,6 +261,12 @@ def _forward_type_inference(self, debug=True):
                     k.replace('output', 'input'): v for k, v in pre_node.output_type.items()
                 }
 
+            # make sure that output nodes have output_type = input_type
+            if isinstance(post_node, Output):
+                post_node.output_type = {
```
Jegp, line 266: Hmmmm this looks like a hack

nir/read.py
```diff
@@ -72,6 +72,7 @@ def read_node(node: typing.Any) -> nir.NIRNode:
             r=node["r"][()],
             v_leak=node["v_leak"][()],
             v_threshold=node["v_threshold"][()],
+            w_in=node["w_in"][()],
```
Jegp, line 75: :skull_and_crossbones:

Jegp (approved):
Hacky, but functional! LGTM :moneybag:

nir/ir.py
stevenabreu7: it feels hacky, but the output node is the only node whose `input_type` field is irrelevant - so it's just a special case where we just have to set the `output_type` as well. a cleaner solution would be that Input and Output only have a `type` - but that would complicate many other things..

Issue #55, opened 2023-09-14
Poor code design pattern for serializing/deserializing


sheiksadique:
At the moment, every primitive that needs to be added requires three different files to be modified. In particular the read and write methods are simply exceptionally long methods with if else conditions. 

The functionality of read write should be moved into the IR definitions so that things are better structured and modularised. This will also make things more scalable.

(We can take inspiration from the memento design pattern)

Jegp:
That's a good point and a good idea. I wonder if we should also somehow split the ir file into smaller parts. If we add these functions, which, I think, is a good idea, it'll grow well beyond it's current ~500 lines  :scream:

sheiksadique:
I would suggest we move each node to its own file. Easier to find and fix things in isolation hopefully.

PR #78, opened 2024-02-29
Reorganise ir and add auto serialisation

alexfanqi:
Hi,

Since it seems no one is working on #55, I took the liberty to fix it. If this happens to be a few hours faster than anyone's work :-P, I am happy to withdraw or discuss for a better implementation.

Deserialization requires a reflection registry to create class from string. I choose to explicitly list all serialisable NIRNodes in a static ir.__all_ir because it is simple and there seems to be no demand for user-defined NIRNode out of the repo. 

Another option would be using `metaclass`/ `__init_subclass__` to dynamically register new sub classes in NIRNode. But I feel `__init_subclass` is black magic.

Would love any feedback for new file structure and implementation.

Cheers


alexfanqi:
Testing was done by first testing new serialization against old deserialization, then testing new {de,}serializations together.

Jegp:
Thank you for the addition! It looks great, but there is one test that's failing because you're using a "|" for a type annotation. Swapping that with [Union](https://docs.python.org/3/library/typing.html#typing.Union) should do the trick. I'm also happy to push the change, if you prefer :)

Jegp:
Thank you for seeing this through @alexfanqi :partying_face:

Jegp:
Closes #55

Issue #70, opened 2023-10-18
meta-data in NIR nodes


stevenabreu7:
In our discussions, the idea came up to include metadata into NIR nodes that would help developers implement NIR nodes on their respective platform. 

As an example, @bvogginger pointed out in #63 that we should not include the reset mechanism into the NIR node definition itself, since the reset mechanism (subtractive, reset to zero, etc) is specific to digital implementations and has no place in the ODE formulation of the node. However, to make sure that NIR graphs commute across platforms, developers need to know what reset mechanism was originally implemented for a given NIR graph. Hence, the idea is to include a file of "metadata" into [every | some] NIR nodes that will give such relevant information. 

Suggestions for metadata:
- LIF/CubaLIF/IF/Threshold nodes: reset mechanism used (e.g. reset-to-zero, subtractive-reset), output spike types used (e.g. graded or not graded - this then also influences the reset mechanism), time step used (for discretization)

sheiksadique:
Thought: A big risk of adding this meta data is libraries getting dependent on this data. This can quickly go downhill from there.

Jegp:
Closed with #79

Jegp:
> Thought: A big risk of adding this meta data is libraries getting dependent on this data. This can quickly go downhill from there.

I added [a warning on this in the docs](https://nnir.readthedocs.io/en/latest/primitives.html#metadata). But yes, this is a big danger we should be concerned and aware about going forward

PR #79, opened 2024-03-03
Added metadata node to all NIRNodes

Jegp:
I added a field with metadata for all NIRNodes for #70 

Note also that this does *not* support nested structures. That's totally by design because I worry this grows out of hand. But it does support arrays.

### DIFF
--- a/nir/serialization.py
+++ b/nir/serialization.py
@@ -4,46 +4,59 @@ from typing import Any, Dict, Union
 import h5py
 import numpy as np
 
+import nir.ir as ir
 import nir
+
+
+def _read_metadata(node: Any) -> Dict[str, Any]:
+    if "metadata" in node.keys():
+        return {"metadata": {k: v[()] for k, v in node["metadata"].items()}}
+    else:
+        return {}
 
 
 def try_byte_to_str(a: Union[bytes, Any]) -> Union[str, Any]:
     return a.decode("utf8") if isinstance(a, bytes) else a
 
 
-def read_node(node: Any) -> nir.typing.NIRNode:
+def read_node(node: Any) -> ir.NIRNode:
     """Read a graph from a HDF/conn5 file."""
     if node["type"][()] == b"Affine":
-        return nir.Affine(weight=node["weight"][()], bias=node["bias"][()])
+        return ir.Affine(
+            weight=node["weight"][()], bias=node["bias"][()], **_read_metadata(node)
+        )
     elif node["type"][()] == b"Conv1d":
-        return nir.Conv1d(
-            input_shape=node["input_shape"][()]
-            if "input_shape" in node.keys()
-            else None,
+        return ir.Conv1d(
+            input_shape=(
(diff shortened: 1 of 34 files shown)


Jegp:
**Update**: The previous version used a feature ([`kw_only`](https://docs.python.org/3/library/dataclasses.html)) that was only supported in Python $\geq$ 3.10. I manually added the fields to all individual objects to avoid this. The downside is that we don't have the field in the abstract `NIRNode` class, but it should work on Python 3.9 which is EOL 2025.

nir/serialization.py
```diff
@@ -4,46 +4,59 @@
...
 def try_byte_to_str(a: Union[bytes, Any]) -> Union[str, Any]:
     return a.decode("utf8") if isinstance(a, bytes) else a
 
 
-def read_node(node: Any) -> nir.typing.NIRNode:
+def read_node(node: Any) -> ir.NIRNode:
     """Read a graph from a HDF/conn5 file."""
     if node["type"][()] == b"Affine":
-        return nir.Affine(weight=node["weight"][()], bias=node["bias"][()])
+        return ir.Affine(
```
matjobst, line 25: Why the change from nir.Affine to ir.Affine? All Node types are directly accessible as nir.Affine, nir.Conv2d etc.

tests/test_readwrite.py
```diff
@@ -40,9 +47,29 @@ def factory_test_graph(ir: nir.NIRGraph):
...
+                assert np.array_equal(v, d2[k])
+            elif isinstance(v, bytes):
+                assert v.decode("utf8") == d2[k]
+            else:
+                assert v == d2[k]
+
+    metadata = {"some": "metadata", "with": 2, "data": np.array([1, 2, 3])}
+    node.metadata = metadata
+    compare_dicts(node.metadata, metadata)
+    tmp = tempfile.mktemp()
```
matjobst, line 63: tempfile.mktemp is actually marked deprecated. They recommend tempfile.mkstemp as more secure variant:
https://docs.python.org/3/library/tempfile.html#tempfile.mktemp

Furthermore, neither mktemp nor mkstemp automatically remove the file. 
I would instead recommend tempfile.TemporaryFile which can be used as context manager, see example here:
https://docs.python.org/3/library/tempfile.html#tempfile-examples
```Python
# create a temporary file using a context manager
with tempfile.TemporaryFile() as fp:
    fp.write(b'Hello world!')
    fp.seek(0)
    fp.read()
# file is now closed and removed
```

nir/serialization.py
Jegp: Good catch! It should indeed be accessed via `nir`

tests/test_readwrite.py
Jegp: Thank you for the input. And, it appears that [h5py also supports temporary files/RawIOBase files as inputs](https://docs.h5py.org/en/stable/high/file.html#python-file-like-objects). I updated the test

PR #86, opened 2024-03-26
fixed NIRNode import issue #85

### LINKED ISSUE #85: Backwards-compatibility of imports broken after refactoring
After the refactoring #78 , the code structure of NIR changed.

Many libraries were using imports such as
```Python
from nir import NIRNode
```

However this fails now as NIRNode is not visible anymore. The solution should be to add it to the list of imports in the `__init__.py` that looks something like this:
```Python
from nir.node import NIRNode
...
__all__ = ["NIRNode", ...]
```

This would allow the simpler imports again.

matjobst:
Fixes #85

### DIFF
--- a/nir/ir/__init__.py
+++ b/nir/ir/__init__.py
@@ -81,6 +81,8 @@ __all__ = [
     "IF",
     "LI",
     "LIF",
+    # node
+    "NIRNode",
     # pooling
     "AvgPool2d",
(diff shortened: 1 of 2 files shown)


bvogginger (approved):
Looks good!

Jegp:
Short and sweet. Thanks!

PR #89, opened 2024-04-10
[Paper] Refactor lava code

stevenabreu7:
Refactoring the lava-based code for the paper experiments.


stevenabreu7:
merged into #90 and continued there

PR #91, opened 2024-04-11
Paper: add CNN pretraining

stevenabreu7:
Adds a notebook for the pre-training of the rate-based CNN, which is then converted to an SNN and used for experiments in the paper.


stevenabreu7:
thank you @biphasic!!

PR #100, opened 2024-04-24
update snntorch example

jeshraghian:
ipynb file has been added. 

It follows the same syntax as the Norse tutorial, though an error is thrown at the following line:

```
# Save to file
nir.write("nir_model.nir", nir_model)
```

Note: I also had to pass in `sample_data` into the network to make it usable, which the Norse example does not have.


Jegp:
There's a type in the "Import a NIR graph to s**n**Torch". Otherwise, LGTM!

stevenabreu7:
fixed, lgtm, merging

Issue #96, opened 2024-04-22
'int' object has no attribute 'item' on conv shapes calculation


chanokin:
This line assumes that the product of the operation/indexing is going to be a tensor/array

https://github.com/neuromorphs/NIR/blob/df2f1fc863557af1c318dfe20f9c3f2b6e5e54db/nir/ir/utils.py#L65

However, I don't know how, I end up with lists here and the elements do not have the `item` method available. You could just check if the `shape` object is an `ndarray` | `tensor` or it has a the `item` attribute before calling it, otherwise just grab the element?

stevenabreu7:
Hi there, thanks a lot for catching this! Could you add a simple test input that gives you this failure? And is the `shape` variable a `list`, or a single `int`? I don't immediately see why it would be a `list`, since we make sure that it would be a `np.array` here:

https://github.com/neuromorphs/NIR/blob/df2f1fc863557af1c318dfe20f9c3f2b6e5e54db/nir/ir/utils.py#L100

stevenabreu7:
I made a PR to fix this, could you please confirm that it works for you now? Then I will merge it. Thanks again for your help!

chanokin:
Hi, sadly, it's still no good. 

Here's a minimal test
```python
from nir.ir.graph import NIRGraph
from nir.ir.conv import Conv2d
from nir.ir.neuron import LI

import torch


def test_nir_conv2d_same():
    # Create a NIR Network
    conv_weights = torch.tensor([[[1.0, 1.0, 1.0], [1.0, 1.0, 1.0], [1.0, 1.0, 1.0]]])
    li_tau = torch.tensor([0.9, 0.8])
    li_r = torch.tensor([1.0, 1.0])
    li_v_leak = torch.tensor([0.0, 0.0])

    nir_network = NIRGraph.from_list(
        Conv2d(
            input_shape=torch.Size((3, 3)),
            weight=conv_weights,
            stride=1,
            padding="same",
            dilation=1,
            groups=1,
            bias=torch.tensor([0.0] * 9),
        ),
        LI(li_tau, li_r, li_v_leak),
    )

```

I think that's because `_index_tuple` still returns a pure `int` in 
https://github.com/neuromorphs/NIR/blob/df2f1fc863557af1c318dfe20f9c3f2b6e5e54db/nir/ir/utils.py#L97-L98

I don't know what the design phylosophy is here but I assume you could return a `np.array([tuple[index]])`

Jegp:
[...]

PR #101, opened 2024-04-28
'int' object has no attribute 'item' on conv shapes calculation (issue 96)

stevenabreu7:
this is a fix for #96.


Jegp (approved):
Looks good :)

PR #103, opened 2024-05-02
Add rockpool example to docs

bauerfe:
I added a example to the docs for conversion between Rockpool and NIR.

However, NIR support in Rockpool has not been publicly released, yet, so there's probably no point yet in merging this. What do you think @Jegp ?


Jegp:
I would agree. But do you know the rough timeline for this?

I'm in the middle of fixing a bug in the Norse code (:grimacing:), so if you're ok with it, I can perhaps add a link that points users in the right direction if they're looking for Rockpool support. Just until a new Rockpool release it out.

bauerfe:
The timeline is currently not clear. All I can say is that hopefully someone will work on it soon.

As a support link, the official  one is rockpool.ai . However, for more direct contact there is also a semi-official discord channel, where people usually reply timely: https://discord.com/channels/852094154188259338/1044914353280921681

stevenabreu7:
LGTM, thanks @DylanMuir @bauerfe @Jegp

PR #107, opened 2024-06-10
Fix to `input_type` node of `CubaLIF` neurons

DylanMuir:
`CubaLIF` neurons would have an `input_type` node of `[]`, instead of `[1]`, when created from a node with scalar `v_threshold`. This pull request fixes this issue.

Also adds a `squeeze()` to the rockpool RNN notebook.


stevenabreu7:
Looks good to me, thanks for the PR @DylanMuir!

DylanMuir:
Hi @stevenabreu7 , it occurred precisely when the neuron population contains 1 neuron. In other cases `v_threshold` was a vector as expected. I did want to check with you about 1D or 2D populations.

The alternative fix would be to ensure `v_threshold` is a vector with a single element, when creating a single neuron.

DylanMuir:
Explicitly, it occurred when exporting a net to nir from `snntorch`, containing `Synaptic` modules.

in `export_nir.py` line 58, in `_extract_snntorch_module()`:

```vthr = module.threshold.detach().numpy()```

The line looks fine, but becomes a scalar numpy array, instead of a single-element vector. So I guess it's up to you whether to fix this in NIR or ask Jason for a fix ;)

DylanMuir:
@stevenabreu7 @Jegp Any thoughts on how to proceed here?

Jegp:
Thank you for following up on this! If I'm understanding this correctly, the scalar seems to be coming from snnTorch. But your addition, @DylanMuir, should work with both scalars and arrays alike, correct? In that case, I think it's an improvement compared to the previous version. Would you agree @stevenabreu7?

Regarding the snnTorch codebase, I opened an issue here: https://github.com/jeshraghian/snntorch/issues/334
I'm not so familiar with the codebase though, so I'm not sure I'm a good candidate to dive in for a fix.

stevenabreu7:
Sorry about the delay.. As it is now, it breaks the types of 2D neuron populations. I pushed a simple change [here](https://github.com/DylanMuir/NIR/pull/1) to ensure that if v_threshold is scalar, it is cast into a numpy array, then we can also keep the `v_treshold.shape` at the end. Does this work for you @DylanMuir?

stevenabreu7:
PR for snnTorch is also up here: https://github.com/jeshraghian/snntorch/pull/335

Jegp:
Can this be merged @stevenabreu7 ? :zombie:  :)

PR #112, opened 2024-08-15
Add metadata to missing nodes

bvogginger:
In the previous version, not all nodes had the `metadata` attribute.
Metadata was now added to all missing nodes.

Also the tests for checking the metadata export/import was updated to consider all nodes in the graph.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.0.277
+  rev: v0.6.0
   hooks:
     - id: ruff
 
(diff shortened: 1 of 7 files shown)


.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.0.277
+  rev: v0.6.0
```
bvogginger, line 4: I updated this to version 0.6.0 to have the same behaviour in pre-commit and during the github-actions-build.

pyproject.toml
```diff
@@ -49,3 +49,5 @@ find={include = ["nir*"]}
 line-length = 100
 lint.per-file-ignores = {"docs/conf.py" = ["E402"]}
 exclude  = ["paper/"]
+extend-exclude = ["*.ipynb"]
```
bvogginger, line 52: With version 0.6.0 of ruff, all notebooks are checked by default in ruff.
This lead to errors when checking the spyx example notebook: https://github.com/neuromorphs/NIR/actions/runs/10415679589.

To restore the previous behaviour, I excluded notebooks again.

stevenabreu7:
Nice, thanks for the PR @bvogginger, it looks good to me!

PR #113, opened 2024-09-03
README typo fixed in 03_rnn

VitF:
In the README for the RNN experiments, there was a mismatch between files and description.


Jegp (approved):
Excellent catch. Thank you for the addition

PR #114, opened 2024-09-20
Renamed surrogate gradient. Added Identity

Jegp:
The threshold functions were previously in the file "surrogate gradients", which struck me as odd since NIR isn't really using surrogate gradients.
I also added an identity node, which I'm using in a separate project.


stevenabreu7 (approved):
LGTM, thanks Jens!

Jegp:
Thanks for the quick review!

Issue #117, opened 2024-11-22
Issues in NIR Documentation


benkroehs:
Hey guys,

I came across some issues in the documentation of NIR that could use attention:
- norse example uses `nir.LI`, which is not implemented in the norse conversion
- lava example uses `nir.Input`'s attribute `shape`, which is not implemented
- links to examples at [this](https://neuroir.org/docs/support.html) page are not working

PR #118, opened 2024-11-22
fix: Issues in examples

benkroehs:
To address the first two issues from #117 , the following code should fix it:


stevenabreu7:
Looks good to me, thanks for the fix!

Issue #106, opened 2024-06-05
Notebook docs run with wrong dependencies


Jegp:
The notebooks in the docs seem to error out because dependencies haven't been installed. E. g. in the Lava notebook where `nir` doesn't exist: https://neuroir.org/docs/examples/lava/nir-conversion.html

Solution: go through the github workflow and ensure that the jupyter book environment has the necessary dependencies installed.

PR #121, opened 2024-12-12
Fix most notebook dependency issues

LuukvanKeeken:
I updated the docs/requirements.txt file, I assume that should help with the dependency issues in the notebooks. Spinnaker2 I couldn't find in pip so that isn't solved here.

There seem to be some other errors in the notebooks as well. Some are just because of incorrect variables (e.g. 'input_type' instead of 'output_type' in the lava example). But also, the way in which the NIRGraph in e.g. the lava example is created doesn't work, seemingly because a dict of nodes is expected instead of a list. Should a different issue be made for these problems?

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -8,3 +8,13 @@ sphinx_external_toc
 sphinxcontrib-mermaid
 astroid~=2.15 # https://github.com/readthedocs/sphinx-autoapi/issues/407
 sphinx-autoapi
+nir
+matplotlib
+lava
+nengo
+norse
+rockpool
+sinabs
+snntorch
+spyx
+jax


Jegp:
Thank you for the changes! @LuukvanKeeken, could I ask you to remind me what this fixes again? I built the docs with your changes and it worked well - but it also worked well without the added dependencies. Of course, the notebooks won't be run, but they still function as static pages.

Is this for interactive notebooks?

LuukvanKeeken:
@Jegp it was regarding this issue: https://github.com/neuromorphs/NIR/issues/106. I never saw the error messages themselves, I assumed they wouldn't be visible to me anyway. Based on the description in the issue I figured that any dependency problems of this kind would be solved by adding the packages to docs/requirements.txt

Jegp:
Ah, right. Thank you! Yes. Ok. I'm afraid the issue is slightly deprecated, which is entirely my fault. The situation is this:
By default the doc builder ([sphinx](https://www.sphinx-doc.org/en/master/)/[jupyter book](https://jupyterbook.org/en/stable/intro.html)) executes the notebook, but we [disabled that earlier](https://github.com/neuromorphs/NIR/blame/main/docs/source/_config.yml#L11) so that Sphinx now converts the notebook content directly into a website without running the content. So, from that perspective, the dependency issue actually disappeared.

But! It's not over. It would be nice to have executable notebooks for two reasons
1. It would permit users to [run the code themselves directly from the docs](https://jupyterbook.org/en/stable/interactive/launchbuttons.html), and 
2. It would be a constant "test" that the code still runs

In some sense this is the original meaning of the issue: we want the notebooks to run, so we can execute them in the docs. Would you be up for working towards that? It requires setting a few configuration options for Jupyter Book and ensuring that the page is generated properly. It's a pretty nice piece of tech that's extremely useful for generating pretty and useful sites.

Sorry for leaving this issue so ambiguous

LuukvanKeeken:
Ah I see! That sounds doable, I'll look into it in the near future

Jegp:
Perfect. Thank you for the understanding @LuukvanKeeken. I'll see to cleaning up in the issues to keep them ajour.

Jegp:
Feel free to open this whenever you find the time @LuukvanKeeken !

Issue #120, opened 2024-12-09
Throw an error when writing/reading empty NIRNodes


Jegp:
NIRNodes are supposed to be abstract and not appear in graphs. When they are instantiated and written/read to/from graphs, we should raise an error.
Currently, we get an empty assertion error if we try to read a serialized graph with NIRNodes.

Jegp:
Closed by #122

Thank you @hanleyc01!

PR #122, opened 2025-01-20
Throw AttributeError for NIRNode constructor

hanleyc01:
# Fix for https://github.com/neuromorphs/NIR/issues/120

- Added inheritance to `NIRNode` from [abc](https://docs.python.org/3/library/abc.html). This might be too much, but given that the stated documentation for `NIRNode` is that it is a "base superclass" and that "NIRNodes should never be instantiated", indicating this property through inheritance in a Pythonic way.
- Throws `AttributeError` on call `NIRNode.__init__` as a proposed fix for [Issue 120](https://github.com/neuromorphs/NIR/issues/120).
- Added a simple test `test_node` which checks if a `NIRNode` class can be constructed. All tests pass, it seems.

# Potential Downsides

- Inheriting from `abc.ABC` might have some unintended consequences in behavior that I am not familiar with, this change I think can be discarded if there are any unintended problems that arise.
- This solution might be a little clumsy and lead to unclear API usage. For example, if a user introspects at runtime on the methods of `NIRNode`, they will find that `__init__` is in fact defined. However, the exception raised here is one used to indicate that there is in fact no `__init__` method. One potential alternative is to raise [`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError), which "may be raised by user code to indicate that an attempted operation on an object is not supported, and is not meant to be".

P.S. This is my first attempted contribution to an open-source library ❤️

### DIFF
--- a/nir/ir/node.py
+++ b/nir/ir/node.py
@@ -1,9 +1,10 @@
+from abc import ABC
 from dataclasses import asdict, dataclass
 from typing import Any, Dict
 
 
 @dataclass(eq=False)
-class NIRNode:
+class NIRNode(ABC):
     """Base superclass of Neural Intermediate Representation Unit (NIR).
 
     All NIR primitives inherit from this class, but NIRNodes should never be
@@ -15,6 +16,9 @@ class NIRNode:
     # input_type: Dict[str, np.ndarray] = field(init=False, kw_only=True)
     # output_type: Dict[str, np.ndarray] = field(init=False, kw_only=True)
     # metadata: Dict[str, Any] = field(init=True, default_factory=dict)
+
+    def __init__(self) -> None:
+        raise AttributeError("NIRNode does not have a default constructor.")
 
     def __eq__(self, other):
         return self is other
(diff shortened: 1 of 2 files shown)


Jegp:
Hi @hanleyc01! Thank you for your pull request (and your patience in our reply :-) )!

This looks great, really. You're correct that this should throw if initialized, because it's not a valid computational primitive in NIR. All tests passed so I'll just go ahead and merge. Congratulations on your first contribution! Don't hesitate to join the discussion on [Discord](https://discord.gg/JRMRGP9h3c) and come back with more changes ;-)

hanleyc01:
> Hi @hanleyc01! Thank you for your pull request (and your patience in our reply :-) )!
> 
> This looks great, really. You're correct that this should throw if initialized, because it's not a valid computational primitive in NIR. All tests passed so I'll just go ahead and merge. Congratulations on your first contribution! Don't hesitate to join the discussion on [Discord](https://discord.gg/JRMRGP9h3c) and come back with more changes ;-)

Thanks so much! I'm glad to have helped, and I'm currently in the discord. I'm preparing for conference submissions with my lab until around the end of this month, but once I get more free time I'll try and get more active in the participation.

PR #124, opened 2025-02-13
Feature check types

Jegp:
Implements @benkroehs type checking default flag with added type checking flags in NIRGraph.from_dict and a VSCode devcontainer


Jegp:
Thanks for seeing this through @benkroehs ! :tada:

PR #125, opened 2025-02-20
General docs improvements

Jegp:
Improves docs by

- Restructuring NIRTorch documentation
- Improving NIR docstrings
- Improving NIR docs on usage and in-depth examples
- Adding API docs for NIR and NIRTorch


docs/requirements.txt
```diff
@@ -7,4 +7,5 @@ jupyter-book
 sphinx_external_toc
 sphinxcontrib-mermaid
 astroid~=2.15 # https://github.com/readthedocs/sphinx-autoapi/issues/407
-sphinx-autoapi
\ No newline at end of file
+sphinx-autoapi
+nirtorch
```
stevenabreu7, line 11: This will make NIR depend on pytorch (bc nirtorch depends on torch), is this necessary? For example, if I'm using a jax-based library, I'd prefer not to have to import torch as well.

stevenabreu7: Perhaps we could have this as a dev_requirement only?

stevenabreu7: now I see that it's only the docs/requirements, so all good!

stevenabreu7 (approved):
This looks great, thanks so much for taking the time to revamp the docs @Jegp!! I have not compiled the docs myself but I trust our CI checks are correct and that you've tested this yourself. Feel free to merge! <3

Jegp:
Thanks for the review @stevenabreu7 <3

PR #126, opened 2025-02-25
Add CuBaLI neuron

benkroehs:
I needed this neuron type as NIRNode as non-spiking Readout Neuron.


Jegp:
Thank you for the addition! Is this supported by a backend or will it be supported by a backend?

I have two requests:
- Can I ask you to add the neuron parameters and equations in the table in the [`primitives.md`](https://github.com/neuromorphs/NIR/blob/main/docs/source/primitives.md) file for documentation?
- Second, can you add a unit test so we can verify that the primitive can serialize and deserialize correctly? It should be straight-forward to copy the [test for the CubaLIF](https://github.com/neuromorphs/NIR/blob/main/tests/test_readwrite.py#L172).

benkroehs:
Ah yes, thank you for mentioning this, I missed this. Both the new table and the test should now be implemented. 
The `CubaLI` is supported by [hxtorch](https://github.com/electronicvisions/hxtorch). We are currently working on NIR support in hxtorch.

Jegp:
Excellent. This looks much better, thank you.
There's an added file `nir/ir/test.py`. Can you remove that? Then I'm happy to proceed with the merge.

benkroehs:
Done, thank you. :)

Jegp:
Thank you for the addition @benkroehs ! :tada:

PR #127, opened 2025-03-05
Update dependencies so tests run smoothly on initial clone

katywarr:
The pytest dependency was missing from the required dependencies so the unit-tests failed after the repo was cloned.
In fixing this, perhaps it's useful to make installation of the project very easy using UV - hence this pull request also proposes a short addition to the README and the incorporation of an associated uv.lock file.

* Add pytest dependency to pyproject.toml
* Add uv.lock file for super-easy installation with uv
* Add a note in the README for newbies to get started quickly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ paper/01_lif/lava
 paper/01_lif/.venv
 paper/02_cnn/lava-dl
 paper/02_cnn/ann_pretraining/data/MNIST
+.DS_Store
--- a/docs/source/contributing.md
+++ b/docs/source/contributing.md
@@ -12,11 +12,7 @@ Use the standard github workflow.
 
 2. Setup the virtual environment for this project.
 
-3. Install all the development requirements.
-
-```shell
-pip install -r dev_requirements.txt
-```
+3. Install all the development requirements (refer to the options described below).
 
 4. Install git pre-commit hooks.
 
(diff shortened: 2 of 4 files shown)


Jegp:
Thank you for the addition. I haven't heard about uv before, but it looks pretty neat.

Currently, your changes would make NIR depending on pytest. I don't think that's wise. That's what the `dev_requirements.txt` is for. So, please remove pytest from pyproject.toml.

I'm happy to add the `uv` addition, but I would request that you move the `uv` description into the documentation. Possibly into the [Using NIR](https://github.com/neuromorphs/NIR/blob/main/docs/source/usage.md), but it may be worth-while adding a page on installation. Of course, that would be a bit more work. Up to you :-)

katywarr:
All done ready for re-review. 

**Documentation:** 
I actually updated `contributing.md` to match the current developer guidance that is already in that file for using `pip`.  It fits best here I think as these are instructions for developers forking the repo, rather than those simply using NIR. 

**Dependencies:**
The development dependencies for uv are identical to those in `dev_requirements.txt` (minus the main dependencies). So, `pyproject.toml` now includes this addition:

```
[dependency-groups]
dev = [
    "black>=25.1.0",
    "pre-commit>=4.1.0",
    "pytest>=8.3.5",
    "ruff>=0.9.9",
    "sphinx>=7.4.7",
]
```

Jegp:
Thank you for the additions! It looks great to me. I'll let the tests run and merge it :)

PR #129, opened 2025-03-17
Add reset potential for IF, LIF and CubaLIF

benkroehs:
With this change I want to propose a `v_reset` attribute to the `IF`, `LIF` and `CubaLIF` node since the reset potential is important for the neurons dynamics. Tell me if there is something missing or if you think things should be implemented in another way. :)


Jegp:
In principle this looks great. You're adding an attribute that's definitely reasonable for many of the models. Simultaneously, this deviates from the neuron models, which means that all the platforms using NIR will need to update their graphs.

I'm not saying it's a bad thing, but I would like to get the opinion of at least a few others of the users. @stevenabreu7 @jeshraghian @DylanMuir @matjobst care to comment?

An additional concern is that all old `.nir` files that's imported with the reset parameters will fail (because the files don't contain `v_reset`). I realize it's not fair to put it on this PR to solve the problem, but before we start changing the primitives, should we maybe consider adding sensible defaults (0)? In that case, we could maintain backwards compatibility. Thoughts @benkroehs ?

stevenabreu7:
If I recall correctly, we had this discussion at some point. Our conclusion was that a threshold subtraction is more suitable for our continuous-time modeling approach. Of course, in continuous-time models, reset and threshold-subtraction are the same. In discrete-time simulations, there will be a mismatch. 

I'm not sure what your goal was for this @benkroehs, but perhaps a good (backward-compatible) compromise would be to introduce a `reset_factor: float` parameter $k$ so that the update equation is something like $v \leftarrow v- k v_{threshold}$ on a spike event instead of $v \leftarrow v_{reset}$ as you propose. Then we can set $k = v_{reset} / v_{threshold}$ to achieve the behavior that you are suggesting. Would this work for your purposes @benkroehs?

benkroehs:
Thank you for your interesting thoughts and concerns.

> Simultaneously, this deviates from the neuron models, which means that all the platforms using NIR will need to update their graphs.

Yes @Jegp, this would change the neuron model. In my view, the current implementation sets the reset potential to zero. To solve this legacy issue, one could set the default value of the reset potential to zero.
Also the different `from_nir` method shouldn't be influenced by this change since it only adds a parameter. 

> An additional concern is that all old `.nir` files that's imported with the reset parameters will fail
That's a good point. As you say, using a default value here too is in my opinion a good way to maintain backward compatibility.

@stevenabreu7 my main goal is to connect [jaxsnn](https://github.com/electronicvisions/jaxsnn/tree/main) as well as [hxtorch](https://github.com/electronicvisions/hxtorch) to NIR. In my current use case, I have to supply the reset potential in a separate conversion configuration for both software frameworks. Although I think the approach with the scaled threshold could work, it doesn't seem very inherent to me since all software frameworks i checked (Norse, snnTorch, jaxsnn, hxtorch) implement a reset potential.

Jegp:
I think @stevenabreu7 brings up an excellent point which I totally overlooked. NIR is inherently continuous, and in a continuous setting the reset is instantaneous. From that perspective it might even be wrong to include this in the IR: what would continuous-time systems do with such a parameter? I don't mean to say we shouldn't have it, but that would be the purist approach.

On the other hand, if the continuous-time systems can't use the parameter, what's the harm in including it? Worst case, it would "just" be a redundant detail.

Another option would be to use the metadata to register this. So, any translation could read the metadata to read what the exact reset mechanism might be. Right now, the metadata is simply a dictionary, so if this is a reasonable approach, we could consider extending the metadata to be more rigorous/typed.
This way, Jaxsnn and hxtorch, could still use NIR to import/export graphs. But they would have to resort to sensible defaults if the data does not exist in the graph (that is, the graph is exported by someone who did not include the reset parameter in the metadata).

Thoughts?

benkroehs:
So in my eyes, regarding that the reset would be instantaneous: Yes, but it also is in the current implementation when the threshold is subtracted from the membrane voltage, no? It only allows this difference to be variable.

And I also do not agree that the reset potential is a redundant part in continuous software frameworks. It is only redundant if you always expect your reset potential to be zero, what is not true in general.

And since the reset potential has the ability to change the neurons dynamics, storing this in the meta data would be incorrect. Also the NIR documentation states "no backend should rely on this metadata".  

So for me the question is _how_ to implement a configurable reset potential. And I think the most inherent way of doing this is to add a parameter `v_reset`, as many software frameworks implemented this too.

DylanMuir:
This is tricky. In continuous time, the two can be made to be the same — by using subtraction, and setting the desired reset value wrt to the threshold.

In discrete time, there is a big difference between the two. Subtraction is computationally preferred, as it loses less information and is a more accurate simulation of the continuous-time model. But some libraries use a fixed reset.

There are many subtle implementation differences between libraries. e.g. order of integration; whether multiple spikes are permitted per dt; which ODE solving approach is used. These shouldn't matter to NIR.

IMO NIR should describe the most flexible general case (which is subtraction of a specified value, IMO), and let the libraries implement as they like.

DylanMuir:
Again, IMO NIR should not *implement* any given neuron model. NIR's value is primarily as an Intermediate Representation. As an aside, I worked with a PhD student yesterday who had built and trained a model in snnTorch, then used NIR to transfer to Rockpool for deployment on Xylo. 12 months ago that would have been a nightmare! It's amazing what NIR is enabling :)

bvogginger:
Let me share my thoughts as well. I think we should separate the discussions:

**1. About adding  a configurable reset voltage/factor**

I fully agree with @benkroehs: The current definition of IF, LIF, CubaLIF does not allow to set the reset voltage. But many backends actually implement it. This PR adds flexibility and it does not hurt existing solutions, as the current behavior can be achieved by setting the reset voltage / factor accordingly. For backwards compatibility, see below.

**2. subtraction vs. reset voltage in continuous time definition**

In my opinion, this is not about continuous-time vs. discrete time vs. implementation. We *have to* add this flexible reset in the *continuous-time formulation*. The question is HOW to formulate it:

@benkroehs 's suggestions:
```
v = v_reset; if spike (v == v_thr)
v = v;           else
```

@stevenabreu7 's suggestion
```
v = v - delta_v_reset; if spike (v == v_thr)
v = v;           else
```
Both are correct and allow to implement the same behavior.

Personally, I prefer the first version, as  `v_reset` is relatable and intuitive. --> It is easier to understand for new users!

Yet, the 2nd option would be more consistent with how we can build *higher-order primitives* (See [Figure 2d of the NIR paper](https://www.nature.com/articles/s41467-024-52259-9/figures/2) from simple primitives. I'm not sure if we can realize the first option (which is actually a short-circuit) by a combination of simple primitive. The second option definitely can be realized.

** 3. Operator versioning **

It will happen that we add new primitives and change existing ones. We should slowly start to think about this versioning and how to support backward compatibility.
We could e.g. look into [ONNX](https://github.com/onnx/onnx/blob/main/docs/Versioning.md) or get inspired by `boost::serialization` which allows to serialize/deserialize objects in different object versions...

benkroehs:
Thank you @bvogginger for this great summary. 

I see that there are different implementations (reset to value vs subtraction). I think that the parameter that is introduced is not necessarily connected to the discussion about reset vs subtraction:
Also a subtraction could be done with a `v_reset` parameter by `delta_v_reset = v_thr - v_reset`. And in my eyes, `v_reset` is a more intuitive parameter than `delta_v_reset` (which would be `v_thr` by default atm).

However, our main objective is to add a parameter representing the reset potential, independent of its final implementation. 
Therefore, I would also be fine with a subtraction. :)

benkroehs:
Hi everyone,

I've now implemented the requested backward compatibility, as suggested by @Jegp. This ensures that "old" .nir files without an explicit reset potential parameter remain compatible.

Let me know what you think – I'd really appreciate your feedback so we can move forward!

benkroehs:
Thank you @bvogginger for the review. 

I agree that this change is not very clean, but in my eyes, it is the cleanest under the  condition that you want to have backward capability.
If we would drop that condition, e.g. the need of default values drops. 
And since the reset potential is a new Parameter, it is also kind of natural that it has to be handled differently - again, as long as we want to address this

bvogginger (approved):
The changes look good to me!

Yet, I'd prefer that this is approved a 2nd person as this is a breaking change.
@Jegp or @stevenabreu7: Could you please give it a short look and let us know if the chosen approach seems appropriate?

Jegp:
Fantastic discussion here. Thanks everyone for mulling this over so carefully. And thank you for adding backwards compatible code @benkroehs!

I like the separation of concerns you made, @bvogginger. Let me address them one by one.

1. **Parameterized reset**

I can see both sides of the coin, and I think it's our job to *permit* both possibilities: parameterizing the reset for discrete simulators that *need* the parameter, but also allow continuous-time systems to *ignore* the parameter. In that sense, I think it's important that we do not hinder either use cases.
Having a separate non-default parameter is annoying. And I'm not sure I follow your argument about "treating parameters equally", Bernhard. We do have other default values for types and weights (in the CubaLIF). I'm actually waiting for [3.9 to reach EOL](https://devguide.python.org/versions/), so we can use [default fields in the initializers](https://github.com/neuromorphs/NIR/blob/main/nir/ir/node.py#L15) (as @benkroehs suggested initially). Why not do that here? Meaning, why not set the parameter to a default value? To me, that would be a little cleaner because we would separate "need-to-have" parameters from "platform-specific" parameters, although I'm not insisting on it.

2. **Subtraction vs hard reset**
Thinking back to the paper experiments, we had to parameterize the reset for different simulators. If you're adamant about adding reset to the equations, this should at least also be accounted for. Otherwise, you'd get different behaviors in different simulators!

Unfortunately, the implementation here is a little gnarly if we want to use a single parameter (as opposed to having an additional string/bool parameter that determines the type of reset). 
As far as I can tell, here are the options:

1. $v \leftarrow v_{reset}$ (doesn't take subtraction into account)
2. $v \leftarrow v - v_{subtract}$ (doesn't allow hard reset)
3. Use the sign of the reset to encode for hard reset/subtraction
  * I really don't like this. It's really obscure and unintuitive.
 
Alternatively, we should include a separate parameter to capture this. Something like

```python
    reset_value: np.array
    reset_method: Literal["reset", "hard"]
```

I prefer the string version over the bool because it's easier to read.

To conclude, I think any implementation now is better than nothing, so I'm happy to merge this and move on. But I think it's a temporary solution and that we should open an issue to deal with the reset types. Otherwise, we're not faithful to the (discrete) integration.

On a more general level, I think we should somehow separate **digital parameters** with **analog parameters**. I agree that we need to represent them somehow, but there's no reason for them to "pollute" the other space. One suggestion for this would be to prefix such parameters with `digital_` and make them all optional. But that probably also requires more discussion.

3. **Operator versioning**
Let's continue that discussion in #136 


**Summary**

I'm happy to merge this as-is, given that we can open an issue to further discuss (1) the integration choices and (2) the naming for digital/analog parameters. Which will be easier to navigate as implementors when we have opset versions in place.

PR #130, opened 2025-04-08
Revamped website

ColonelParrot:
Redesigned the home page and added a credits page. Production ready and SEO compliant!

You can see both in the wild here:

Home page: https://colonelparrot.github.io/nir-website-beta/
Credits page: https://colonelparrot.github.io/nir-website-beta/credits


Jegp:
Super nice work @ColonelParrot! It runs smoothly, in all sorts of browsers and with all sorts of mobile screen resolutions.

I just have two comments: 
- the figures in the "Why NIR" section need a bit of revamping. Some of them are a little simple. Could I be helpful here?
- could you reduce the screen height a little in the "Why NIR" section so people can see that there is more stuff coming below?

Jegp:
Let's merge this into the `feature-web` branch and continue from there!

PR #140, opened 2025-06-09
Infer output types

fabio-innatera:
Set `output_type` of NIRGraph and Output node `Output.input_type` when doing forward type inference.

I think this is fine if the output_type is not set yet, but overriding might be problematic? Unsure.

My thinking is that if we do forward type inference from the input types, we should assume the output_type of the Output node and NIRGraph are either not set or wrong, therefore we should update them.

Immediate use-case for me is updating the Output node type when importing NIR into PyTorch: https://github.com/neuromorphs/NIRTorch/blob/c5b101532c3d8f549b10323270fa13dae461b402/nirtorch/torch_tracer.py#L167

Also, this should be useful if we implement type inference on nested graphs, for that we need valid input_type/output_type on the nested graphs.

This PR is based on top of #139, and the one should be merged first.


Jegp:
> > Great addition, thank you. Both the typing of the `Output` node and the graph itself is valuable. Couldn't one argue that the same thing should happen for the `input_type` of the `NIRGraph`?
> 
> `_forward_type_inference` is not called by `infer_types` if any of the entries in `NIRGraph.input_type` is `None`. `NIRGraph.input_type` is set to the input_types of the `Input` nodes in `NIRGraph.__post_init`. This means I can't see a case where there is actually anything to update.

Ah, I see. Could it be that the job of the type inference is not only to update it, but also to check it? That is, complain if any of the types are malformed?

> > I would request two small changes:
> > 
> > 1. It's a good idea to notify the user if the output type is overwritten, but I wouldn't use `print` in the first case. What about using something like [`warnings.warn`](https://docs.python.org/3/library/warnings.html)? If people want to direct the output somewhere else, they can do that with a simple [`with` statement](https://docs.python.org/3/library/warnings.html#overriding-the-default-filter).
> 
> I am unsure what the right solution is here. For example, for this line, I don't think we should print anything at all, this is expected behavior from type inference: `f"[warning] {post_key}.input_type undefined, set to {pre_key}.output_type"`
> 
> I am not a big fan of the warnings module. One annoying thing for example is that warnings show up in pytest output. We already have a `debug` argument for the `_forward_type_inference`, maybe we should use that? And also add it to `infer_types`?
> 
> This way, if the user uses infer_types interactively, they will see the warnings, but for library users (e.g. in NIRTorch), we can disable it.
> 
> Maybe the logging module would be more appropriate? Log it with level debug? But also happy to go with warnings if there are good arguments for it.

My understanding is that warnings can be made to comply with the logging module (https://stackoverflow.com/a/37979724), but I agree with your point that we should stick with one logging system. I'm happy to go with `logging` over `warnings` given your argument.

We could use the `debug` argument, but if we build on the `logging` module, and we're only printing stuff anyway, shouldn't we use their config?
The argument is btw also unused at the moment which is a little scary.

> 
> > 2. Run `ruff` on the code to ensure the formatting is correct. The linting test is failing.
> 
> Done. I also opened a PR #147 to check the actual code formatting with black on CI, right now we only do linting checks.
> 

Brilliant, thanks!

> > Oh, and thanks for catching the double implementation of `_forward_type_inference`!
> 
> Happy to help improve the codebase :)

<3

fabio-innatera:
This is superseded by #153, so I will close this PR.

Due to how everything works now in that PR, the conclusion to the warning discussion for now is that we do not warn - if an input is wrong, this is an error. Somebody made a mistake constructing the graph somewhere, and they should fix it.

Issue #142, opened 2025-06-11
New reset potential on IF, LIF, CubaLIF breaks existing NIR files


fabio-innatera:
Loading an existing *.nir file containing one of those three primitives exported by an older NIR version leads to a type error, because the serialized dict does not contain the `v_reset` dictionary key. This happens with the latest NIR version from GitHub, the one released on PyPI is not affected.

I think we could either
* Fix the importer to take care of this
* Make v_reset a default parameter

For example, trying to load a file from the paper https://github.com/neuromorphs/NIR/blob/6189cacd1b1e89d1e587afd90f0f8930361bae98/paper/03_rnn/braille_noDelay_noBias_subtract.nir gives the following error:

```python
import nir
nir.read("braille_noDelay_noBias_subtract.nir")
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], line 2
      1 import nir
----> 2 nir.read("braille_noDelay_noBias_subtract.nir")

File /workspaces/nir-dev/NIR/nir/serialization.py:181, in read(filename)
    179 with h5py.File(filename, "r") as f:
    180     data_dict = hdf2dict(f["node"])
--> 181     return nir.dict2NIRNode(data_dict)

File /workspaces/nir-dev/NIR/nir/ir/__init__.py:58, in dict2NIRNode(data_dict)
     51 def dict2NIRNode(data_dict: Dict[str, Any]) -> NIRNode:
     52     """Assume data_dict["type"] exist and correspond to a subclass of NIRNode.
     53 
     54     Other items should match fields in the corresponding NIRNode subclass, unless
     55     subclass provides from_dict. Any extra item will be rejected and should be removed
     56     before calling this function
     57     """
---> 58     return str2NIRNode(data_dict["type"]).from_dict(data_dict)

File /workspaces/nir-dev/NIR/nir/ir/graph.py:160, in NIRGraph.from_dict(cls, kwargs)
    156     f"{type}. Either remove the entry or use <Specific NIRNode>.from_dict, such as Input.from_dict"
    157 kwargs_local["type"] = "NIRGraph"
--> 160 kwargs_local["nodes"] = {k: dict2NIRNode(n) for k, n in kwargs_local["nodes"].items()}
    161 # h5py deserializes edges into a numpy array of type bytes and dtype=object,
    162 # hence using ensure_str here
[...]

PR #143, opened 2025-06-11
Fix deserialization of legacy .nir files

benkroehs:
This PR fixes the broken import of .nir files which were created before #129, mentioned in #142.


fabio-innatera:
Small nitpick, otherwise looking good :+1:

PR #147, opened 2025-06-13
CI: Check code formatting

fabio-innatera:
I guess we could also discuss doing both linting + code formatting with ruff, but that is something for another time I think.


Jegp:
Thanks for this.

We do, by the way, have a [pre-commit-config](https://github.com/neuromorphs/NIR/blob/main/.pre-commit-config.yaml) that should do this automatically.

PR #148, opened 2025-06-16
Prevent from_dict from modifying non-local scope

mrontio:
kwargs here is a pointer, so del would delete it from the parent variable.


Jegp (approved):
Really nice catch here. Thanks for the addition!

PR #152, opened 2025-06-16
Fix black to 25.1.0 in pre-commit too

mrontio:
Extension to mrontio:black-format-fix pull request


mrontio:
Apologies, my mistake for not including this in the original request.

PR #153, opened 2025-06-19
Enhanced type inference and checking, called from NIRGraph.__init__

fabio-innatera:
This is my first stab at doing type inference, in addition to type checking, on graph construction (`NIRGraph.__init__`).

There are still a few things to do, so I put this up as a draft for now, but I think it would be a good idea for
people to have a first look and comment if this is going in the right direction.

All commits should have passing tests, so we can pick-and-chose if necessary

# Input and Output node handling
We automatically add Input and Output nodes if they are missing from the NIRGraph. Before, we only did this on NIRGraph.from_list, but we now do this on all graphs. This is a requirement for nested graphs, otherwise we can't validate the outputs from nested graphs against the data going in and out of it from the outer graph.

A major break is the requirement of Input and Output nodes to be present for graphs that have no non-ambiguous  inputs or outputs. This is the case for certain recurrent connections, for example:
```mermaid
graph LR
A --> B
B --> A
```
For this graph, it's impossible to know which one is the input/source node and which one is the output/sink node, so the type inference will throw an error.
This graph with a recurrent connection is fine:
```mermaid
graph LR
input --> A
A --> B
B --> A
A --> output
```

I think this change limits the flexibility of NIR in some way, at least when creating graphs with type inference/checking active, so needs careful thought.

TODO:
* [ ] Support nested graphs for both type checking and type inference
* [ ] Reactivate more currently skipped tests, that should be fixed with this change.
* [ ] Add some more specific tests for type inference

### DIFF
--- a/nir/ir/graph.py
+++ b/nir/ir/graph.py
@@ -55,7 +55,7 @@ class NIRGraph(NIRNode):
 
         # Check that all nodes have input and output types, if requested (default)
         if type_check:
-            self._check_types()
+            self.check_types()
 
         # Call post init to set input_type and output_type
         self.__post_init__()
@@ -192,7 +192,7 @@ class NIRGraph(NIRNode):
         else:
             raise ValueError("Either input_type or output_type must be set")
 
-    def _check_types(self):
+    def check_types(self):
         """Check that all nodes in the graph have input and output types.
 
         Will raise ValueError if any node has no input or output type, or if the types
--- a/nir/ir/graph.py
+++ b/nir/ir/graph.py
@@ -416,6 +416,14 @@ class NIRGraph(NIRNode):
         source_nodes = {edge[0] for edge in self.edges}
         leaf_nodes = all_node_keys - source_nodes
 
+        if not leaf_nodes:
+            raise ValueError(
+                "Type inference failed: No output nodes found. "
+                "This may be due to a cyclic dependency at the graph's output. "
+                "Please add an `Output` node manually to define an exit point, "
+                "or disable type checking (`type_check=False`)."
+            )
+
         new_nodes: Dict[str, NIRNode] = {}
         new_edges: Edges = []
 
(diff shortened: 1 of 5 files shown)


nir/ir/graph.py
```diff
@@ -234,7 +242,7 @@ def _check_types(self):
                 )
         return True
 
-    def _forward_type_inference(self, debug=True):
+    def _forward_type_inference(self):
```
Jegp, line 245: Should this be a non-hidden method (without the initial `_`)? I'm asking because it may be that someone wants to actually run this on graphs independently of the constructor.

If yes, what about going with an imperative verb-construct such as `infer_types` similar to `check_types`?

Jegp (approved):
Excellent work! I really appreciate the thought you put into this and the number of tests you added to cover this.

I have one comment around the naming for `_forward_type_inference` that I'd like your opinion on. Apart from that, I couldn't find any problems. Nice!

You bring a good point that this may mess up old graphs/weirdly shaped graphs. But you neatly wrapped the inference + type checking in the constructor, so I still think it should be possible to build the graphs and ignore the type checking. I retain that the checking should be default and that any choice to ignore the typing should be a conscious choice. So, I honestly like this model.

nir/ir/graph.py
fabio-innatera: We already have `infer_types` that dispatched either to `_forward_type_inference` or the non-existing backward type inference.

I would like to propose to remove this dispatch function and rename `_forward_type_inference` to `infer_types`. And also move the checks that are currently in `infer_types` to the to-be-renamed `_forward_type_inference`.

Or do you think backward-mode type inference will be added anytime soon?

Jegp: Fully agree with your proposal to merge the two functions. As you say, `infer_types` seems more or less just like a wrapper, and merging them would give much cleaner code I think. Do you want to do that in this PR or should we shift that to the future?

fabio-innatera: I added it to this PR.

PR #157, opened 2025-07-21
Support scalar parameters in type inference

fabio-innatera:
I am trying to import a NIR file exported by latest snnTorch from GitHub, and got an error in type inference.

Turns out that snnTorch uses numpy scalars like `np.float32(4.0)` to define parameters, which end up
as `input_type={'input': ()}` (empty tuple). This implies that the actual neuron count is not known on the neuron node.

Technically, snnTorch is wrong here, the type annotation for neurons call for a np.array, but because numpy scalar values like `np.float32` have a `.shape` property, this actually worked just fine before we introduced type inference.

Thankfully we are already doing type inference, and in most cases we can infer the number of neurons and therefore input_type and output_type from preceding nodes, and this is what I am doing here.

### DIFF
--- a/nir/ir/graph.py
+++ b/nir/ir/graph.py
@@ -6,6 +6,7 @@ import numpy as np
 
 from .conv import Conv1d, Conv2d
 from .flatten import Flatten
+from .neuron import IF, LI, LIF, I
 from .node import NIRNode
 from .pooling import AvgPool2d, SumPool2d
 from .typing import Edges, Nodes, Types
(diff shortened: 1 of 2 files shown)


fabio-innatera:
Maybe we should also change the type annotations?

```python
import typing
import numpy.typing as npt

typing.Union[np.floating, npt.NDArray[np.floating]]
```

fabio-innatera:
I agree that we should enforce usage of np.array instead of loosening the constraints.

I opened a new PR that does just that: #162 

Digging around in snntorch, it looks like at least the tests already use array parameters: https://github.com/jeshraghian/snntorch/blob/master/tests/test_nir.py#L60

PR #159, opened 2025-07-24
Add overview of supported NIR nodes per framework

benkroehs:
### Motivation
With this addition, I aim to improve the NIR documentation by clearly indicating which NIRNodes are supported by which software frameworks. This has two key benefits:

1. Developers can easily see which Nodes are already supported in other frameworks, helping them identify which ones might be valuable to support in their own projects.
2. Users can quickly find out which frameworks support the Nodes they need, without having to dig through GitHub repositories to gather this information manually.

### State of the PR

First of all, this is just a draft. There are still some issues (see below) and I'm hoping to get feedback on it.
The main idea is that I didn't want to manually update this list every few months whenever a framework starts supporting a new NIRNode. So I thought it would be useful to make this process dynamic.
To achieve that, I manually specify the URLs of all the `from_nir.py` and `to_nir.py` fiels and lable them accordingly. 
The hackiest part is probably the detection logic: I search these files for `"ir.{node}:"`, `"ir.{node}("`, or `"ir.{node})"` for each `node` in a list of existing NIRNodes. (I can't omit the :, (, ) because otherwise, `nir.IF` will also be detected as `nir.I`)
This information is then formatted in a table
But yeah, while this mostly works, there are some problems with the current implementation.

### Issues

1. There are frameworks which to not have two distinct files for both directions of the NIR conversion. These are rockpool, sinabs and Spyx. I'm not sure how to handle this. Maybe the query can split the file in a from NIR and to NIR part. 
2. Frameworks (like jaxsnn and Spyx) mention NIRNodes in the conversion which they do not support. I will address this for jaxsnn but actually it is no problem to mention these nodes there. Only my way of creating this table does not work with it.
3. Finally I am not sure if there is a much easier way then just searching through the files. Or does it even make sense to do the whole thing automatically?

I would really like to hear some thoughts on this. :)

(This also depends on #158)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 [![Discord](https://img.shields.io/discord/1209533869733453844?logo=discord)](https://discord.gg/JRMRGP9h3c)
 
 NIR is a set of computational primitives, shared across different neuromorphic frameworks and technology stacks.
-**NIR is currently supported by 7 simulators and 4 hardware platforms**, allowing users to seamlessly move between any of these platforms.
+**NIR is currently supported by 8 simulators and 5 hardware platforms**, allowing users to seamlessly move between any of these platforms.
 
 NIR is useful when you want to move a model from one platform to another, for instance from a simulator to a hardware platform.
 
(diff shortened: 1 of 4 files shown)


Jegp (commented):
I think the idea to automate this is excellent. None of us have the time to update each entry and/or monitor each library independently. But I'm not sure grepping for `ir.{node}` is a good idea. What if people do `from nir import LIF` and then straight-up uses `LIF`?

Come to think of it, one way to test for the support of a node is to actually RUN it through the framework. What if we rewrote the script to (1) for each NIR node, create a NIR graph with a single node that we then (2) try to input to the library and (3) export again. Any failure would mean that the node wasn't supported. The downside of this is that cannot detect an export feature if the import doesn't work. Anyway, just an idea. I'm sure we can also tweak the text-grepping method to be more robust.

Also, I'd add a GH workflow script to update this once in a while (doesn't have to be often).

Btw, Lava and Nengo never merged their code upstream. It exists, but it's not publicly supported.

Jegp:
> I am not sure about your proposal of converting the NIR Nodes to the framework and back again. I have two concerns
> 
>     1. So as far as I can see, the nodes a network supports for import from NIR and export to NIR might differ. So if a Node is only supported for to NIR conversion, this won't be detectable (this is at least once the case: Linear in Norse)

That's an excellent point. The map isn't bijective, so this won't work.

> But I am thankful for every thought on this, so I think we can try this and maybe this works better. :)

Let's stick with the grepping thing for now. The remaining task will then be to 
1. Figure out how to parse frameworks that don't follow the typical to/from structure.
2. Write some workflow that periodically updates the support matrix.

Agreed?

An independent thought would be to actually dump in the supported nodes in the [individual platform support pages](https://neuroir.org/docs/examples/jaxsnn/nir-conversion.html). It'd be nice to know what jaxsnn supports when you read about it, for instance.

bvogginger:
Note: For SpiNNaker2 we list the supported NIR primitives in the pages: https://spinnaker2.gitlab.io/py-spinnaker2/user_guide/qs-nir.html#supported-nir-primitives

PR #162, opened 2025-08-01
raise exception when user tries to create neurons with scalar parameters

fabio-innatera:
I am proposing this as an alternative to #157. The user should get a useful error message instead of a cryptic error message inside type inference.

Not sure about the exact wording for the error message? Is CubaLIF technically (just) a neuron?

### DIFF
--- a/nir/ir/neuron.py
+++ b/nir/ir/neuron.py
@@ -43,6 +43,9 @@ class CubaLI(NIRNode):
             == self.r.shape
             == self.v_leak.shape
         ), "All parameters must have the same shape"
+        assert (
+            len(self.tau_syn.shape) != 0
+        ), "Neuron parameters cannot be scalar, only numpy arrays"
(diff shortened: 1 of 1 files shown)


Jegp:
Hmmm I agree that it's broken in snnTorch, but that's not necessarily the case for other simulators. Isn't it still possible to define scalar parameters if broadcasting is supported? I'm just wondering whether we should save other platforms from the same incompatibility troubles that snnTorch went through :)

Jegp:
I like the idea of emitting warnings, although we should be careful not to spam too much. And someone did raise a good point about warnings not being optimal, although I can't find the comment now. Does that ring a bell @fabio-innatera ?

I think the path of least resistance is to (1) try to fix graphs as best we can to maintain backward compatilibity, but (2) let the user know that they're entering dangerous territory. We can introduce deprecation annotations and then be more strict in future versions so people have time to adjust. What do you think?

PR #164, opened 2025-08-14
Improved LAVA <> NIR interface 

alejandroggener:
Created from the pull request in lava-dl [lava_to_nir](https://github.com/lava-nc/lava-dl/pull/218/files#diff-b8d114c7398f12e3826411a8227dd0466dbf0f41b14dfdcc8e50db0adc4c6676) and the official documentation of NIR and the [lava_to_nir](https://github.com/neuromorphs/NIR/blob/main/paper/nir_to_lava.py#L67) python script.

Modified two lib scripts `lava_to_nir.py` and `nir_to_lava.py` that have been tested for the [Oxford example in lava-dl](https://lava-nc.org/lava-lib-dl/slayer/notebooks/oxford/train.html) 

For testing purposes, use the oxford_training and its data in [Oxford example in lava-dl](https://lava-nc.org/lava-lib-dl/slayer/notebooks/oxford/train.html) 

 ## Main modifications `lava_to_nir.py`

 1. Lava-dl only uses one size parameters for all neurons. Exported to NIR as numpy arrays with N data according to number of neurons per layer
2. Saving floating point parameters of CUBA LiF neurons in NIR instead of hardware fixed point values.
3. Added dt value when exporting from LAVA-dl to NIR.

## Main modifications `nir_to_lava.py`
 1. Introduced dynamic n_neurons per layer.

## Main limitations

1. Considering `dt=1e-4` according to `nir_to_lava.py` script. Need to check this with LAVA developers.

2. Tested only 1D Dense CUBA LiF Neurons. According to the code, there is more work to do for 2D Dense layers. 

3. Didn't test different Neurons. 

## CUBA LiF. LAVA to NIR

Here it is a small piece of code that represents the matching of different values from LAVA-dl to NIR. Double check if this is correct with NIR developers. 

```
# Lava-dl exports hardware fixed tipe to hdf5. For NIR, export in floating point. 
        dt      = 1e-4  # This dt value is according to nir_to_lava.py script. https://github.com/neuromorphs/NIR/blob/main/paper/nir_to_lava.py#L67
        vdecay  = layer['neuron']['vDecay'][()]  / 4096 # Save the value in NIR as floating point
        idecay  = layer['neuron']['iDecay'][()]  / 4096 # Save the value in NIR as floating point
        thr     = layer['neuron']['vThMant'][()] / 64   # Save the value in NIR as floating point
        tau_mem = dt/float(vdecay) if vdecay != 0 else np.inf
        tau_syn = dt/float(idecay) if idecay != 0 else np.inf
        shape   = layer['weight'].shape[0]
        r       = tau_mem/dt # no scaling of synaptic current
        w_in    = tau_syn/dt # no scaling of synaptic voltage

```

### DIFF
--- a/paper/lava_to_nir.py
+++ b/paper/lava_to_nir.py
@@ -1,0 +1,274 @@
+import typing
+import h5py
+import re
+import pathlib
+import nir
+import numpy as np
+import logging
+from typing import Tuple
+import collections.abc
+
+
+PATH_TYPE = typing.Union[str, pathlib.Path]
+
+
+def read_cuba_lif(layer: h5py.Group, shape: Tuple[int] = None) -> nir.NIRNode:
+    """Reads a CUBA LIF layer from a h5py.Group.
+    TODOs: 
+    - what if the layer is more than 1D?
+    - handle scaleRho tauRho theta
+    - support graded spikes
+    - support refdelay
+    - support other neuron types
+    If the neuron model is not supported, a warning is logged and None is returned.
+    """
+    logging.debug(f"read_cuba_lif {layer['neuron']['type'][()]}")
+
+    if 'gradedSpike' in layer['neuron']:
+        if layer['neuron']['gradedSpike'][()]:
+            logging.warning('graded spikes not supported')
+
+    if layer['neuron']['type'][()] in [b'LOIHI', b'CUBA']:
+        if layer['neuron']['refDelay'][()] != 1:
+            logging.warning('refdelay not supported, setting to 1')
+        if layer['neuron']['vDecay'] == 0:
+            logging.warning('vDecay is 0, setting to inf')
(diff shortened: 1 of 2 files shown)


alejandroggener:
hi @Jegp. Sounds like a plan! I can prepare an extra notebook with the whole Oxford example working with NIR, to parse the network to NIR and LAVA again, so we have a whole SNN as an example. Two main questions.

1. Do you prefer to modify this pull request or add a new one with the changes?

2. Should I also copy lava_to_nir.py and nir_to_lava.py scripts into the [https://github.com/neuromorphs/NIR/blob/main/docs/source/examples/lava/](https://github.com/neuromorphs/NIR/blob/main/docs/source/examples/lava/) directory?

PR #166, opened 2025-10-01
Added some units to the primitives

Jegp:


Jegp:
@rowleya could I ask for your quick input here? Does this clarify some of the units?

I realize we're not putting units on everything, but it's a start :)

PR #175, opened 2025-11-24
feat: Add Intermediate Representation for Data in SNNs

benkroehs:
This initial PR introduces NIRData, an intermediate format for data representations in spiking neural networks.
Its purpose is to streamline data exchange across existing simulators and neuromorphic hardware, which currently rely on incompatible formats—typically either time-gridded tensors of 0s and 1s or event-based as lists of neuron indeces and event times.

NIRData mirrors the structure of NIR itself:
* NIRGraphData organizes the NIRNodeData objects
* NIRNodeData provides node-level observables (e.g., spikes, current, …).

Conversion to and from this format is intended to be implemented by the respective software/hardware platforms. As an example, these conversions are is implemented for two software frameworks of our group: The event-based conversion is implemented in [jaxsnn](https://github.com/electronicvisions/jaxsnn/blob/main/src/pyjaxsnn/jaxsnn/event/to_nir_data.py), and the time-gridded one in [hxtorch](https://github.com/electronicvisions/hxtorch/blob/master/src/pyhxtorch/hxtorch/spiking/utils/to_nir_data.py).  (there is each a `from_nir_data.py` and a `to_nir_data.py` file)

The overall goal is to take a step toward unifying data formats, increasing accessibility and improving interoperability across existing simulators and hardware platforms.

What are your thoughts on this?

CC @muffgaga

### DIFF
--- a/nir/__init__.py
+++ b/nir/__init__.py
@@ -16,6 +16,6 @@ from . import ir, data_ir
 from .ir import *  # noqa: F403
 from .ir import typing  # noqa: F401
 from .data_ir import *  # noqa: F403
-from .serialization import read, write
+from .serialization import read, write, read_data, write_data
 
-__all__ = ir.__all__ + data_ir.__all__ + ["read", "write"]
+__all__ = ir.__all__ + data_ir.__all__ + ["read", "write", "read_data", "write_data"]
--- a/nir/serialization.py
+++ b/nir/serialization.py
@@ -258,3 +258,170 @@ def write(
         f.create_dataset("version", data=nir.version, dtype=h5py.string_dtype())
         node_group = f.create_group("node")
         write_recursive(node_group, graph.to_dict())
+
+
+def read_data(path: str) -> nir.NIRGraphData:
+    """Load NIRGraphData from a HDF/conn5 file.
+
+    Arguments:
+        filename (Union[str, Path]): The filename as either a string or pathlib Path.
+
+    Returns:
+        NIRGraphData read from the file.
+    """
+
+    def _read_time_gridded(group: h5py.Group) -> nir.TimeGriddedData:
+        data = group["data"][()]
+        dt = float(group.attrs["dt"])
+        return nir.TimeGriddedData(data=data, dt=dt)
+
+    def _read_event_data(group: h5py.Group) -> nir.EventData:
+        idx = group["idx"][()]
+        time = group["time"][()]
+        n_neurons = int(group.attrs["n_neurons"])
+        t_max = float(group.attrs["t_max"])
+
+        if "__type__" in group.attrs and group.attrs["__type__"] == "ValuedEventData":
+            value = group["value"][()]
+            return nir.ValuedEventData(
+                idx=idx,
+                time=time,
+                value=value,
+                n_neurons=n_neurons,
+                t_max=t_max,
+            )
+
+        return nir.EventData(
+            idx=idx,
+            time=time,
+            n_neurons=n_neurons,
+            t_max=t_max,
+        )
+
+    def _read_node_data(group: h5py.Group) -> nir.NIRNodeData:
+        observables = {}
+        obs_group = group["observables"]
+
+        for name, g in obs_group.items():
+            typ = g.attrs["__type__"]
+            if typ == "TimeGriddedData":
+                observables[name] = _read_time_gridded(g)
+            elif typ in ("EventData", "ValuedEventData"):
+                observables[name] = _read_event_data(g)
+            else:
+                raise ValueError(f"Unknown observable type: {typ}")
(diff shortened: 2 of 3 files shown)


INSTALL.md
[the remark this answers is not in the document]
Jegp: Or maybe the CI/CD?

nir/data_ir/graph.py
```diff
@@ -0,0 +1,261 @@
...
+
+@dataclass
+class ValuedEventData(EventData):
+    """
+    Valued event-based data as a list of event indices, event times and event
+    values.
+
+    Parameters
+    ----------
+    idx : np.ndarray[int], shape (n_samples, n_events)
```
Jegp: These parameters don't seem to align with the constructor of the dataclass (`values`). In fact, where are the `idx` and `time` class variables?! Is the `test_generate_valued_event_data` test passing?

nir/data_ir/graph.py
```diff
@@ -0,0 +1,261 @@
+from __future__ import annotations
+from dataclasses import dataclass
+from enum import Enum
+from typing import Dict, Union
+import numpy as np
+from nir import NIRGraph
+
+
+class Interpolation(Enum):
+    NONE = "None"
```
Jegp: Why should this be an Enum? Do we expect other kinds of interpolation methods in the future?

tests/test_data_ir.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
+import nir
+
+
+def test_generate_time_gridded_data():
+    spikes = np.random.randint(0, 2, size=(10, 10, 10)).astype(bool)
+    dt = 0.1
+    gridded = nir.TimeGriddedData(spikes, dt)
+    node = nir.NIRNodeData({"spikes": gridded})
+    graph = nir.NIRGraphData({"node": node})  # noqa: F841
```
Jegp: It's great to test that you can create the objects, but it would be good to also verify the content. A simple test could be to ensure the information is preserved. It's basic, but something like `np.allclose(graph.nodes["node"].data, spikes)` would make sure we won't have unexpected regressions.

tests/test_data_ir.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_generate_valued_event_data():
+    idx = np.random.randint(0, 10, size=(10, 20))
+    time = np.sort(np.random.rand(10, 20), axis=1)
+    values = np.random.rand(10, 20)
+    n_neurons = 10
+    t_max = 1.0
+    valued_event = nir.ValuedEventData(idx, time, n_neurons, t_max, values)
+    node = nir.NIRNodeData({"current": valued_event})
+    graph = nir.NIRGraphData({"node": node})  # noqa: F841
```
Jegp: Same.

tests/test_data_ir.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def test_binary_conversion():
+    # time_shift = 0.0 * dt
+    spikes = np.random.randint(0, 2, size=(10, 10, 10)).astype(bool)
+    dt = 0.1
+    gridded_1 = nir.TimeGriddedData(spikes, dt)
+    event = gridded_1.to_event(n_events=100)
+    gridded_2 = event.to_time_gridded(dt=dt)
+    assert np.array_equal(gridded_1.data, gridded_2.data)
```
Jegp: Nice! :-)

nir/__init__.py
```diff
@@ -15,6 +15,7 @@
 from . import ir
 from .ir import *  # noqa: F403
 from .ir import typing  # noqa: F401
+from .data_ir import *  # noqa: F403
```
Jegp: We should also add it to the `__all__` list, so it features when people import nir.

`__all__ = ir.__all__ + data_ir.__all__ + ["read", "write"]` should do it ( if `__all__` is defined in `data_ir/__init__.py`)

INSTALL.md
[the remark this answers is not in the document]
benkroehs: Ah, I forgot to delete this. This was part of our local CI.

nir/data_ir/graph.py
```diff
@@ -0,0 +1,261 @@
...
+
+    observables: Dict[str, Union[EventData, TimeGriddedData]]
+
+    def __post_init__(self):
+        if not isinstance(self.observables, dict):
+            raise TypeError(
+                "observables must be a dictionary of EventData or TimeGriddedData"
+            )
+
+    def check_observables():  # TODO: should be called by check nodes
```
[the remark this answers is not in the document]
benkroehs: done

Jegp:
Oh yes, one more thing. Put this in a paper @benkroehs ! :P

PR #187, opened 2026-02-25
fix: use dict key lookup instead of hasattr in read()

siddiquifaras:
The type_check guard in read() used hasattr(data_dict, "type_check") but data_dict is a dict, so this checks for an attribute on the object rather than a key in the dict. It is always False and the guard never fires. Changed to "type_check" in data_dict.


Jegp (approved):
Ah, great catch!

siddiquifaras:
Thank you

PR #188, opened 2026-02-25
fix: add structural graph validation before type inference

siddiquifaras:
Adds validate_structure() to catch dangling edges and duplicate edges at the start of infer_types(). Previously these surfaced as confusing type inference failures deep in the traversal. Now they produce clear error messages naming the bad edge and listing available nodes.


Jegp (approved):
Very nice work. The code looks clean and you added tests. Thanks!

siddiquifaras:
Whoops, a check failed... I'll have a look at fixing that

PR #189, opened 2026-02-25
fix: resolve scalar parameter types from graph context during inference

### LINKED ISSUE #176: Fix: Scalar Parameter Broadcasting Forces Inefficient Weight Duplication
## Summary

NIR's type inference system currently requires neuron parameters (tau, v_threshold, etc.) to be arrays with explicit shapes to determine the number of neurons in a layer. This forces scalar parameters to be broadcast into full arrays, creating unnecessary weight duplication that is inefficient for hardware deployment.

## Problem Description

When exporting spiking neural networks to NIR, scalar parameters (which are common and hardware-efficient) must be converted to full arrays to satisfy NIR's type checking system.

### Example

**Original snnTorch model:**
```python
import snntorch as snn
lif = snn.Leaky(beta=0.9)  # Single scalar parameter for all neurons
```

**Current NIR export requirement:**
```python
# For a 128-neuron layer, scalar beta=0.9 becomes:
nir.LIF(
    tau=np.array([0.001, 0.001, 0.001, ..., 0.001]),  # 128 identical values
    v_threshold=np.array([1.0, 1.0, 1.0, ..., 1.0]),  # 128 identical values
    # ... all parameters duplicated 128 times
)
```

**What we want:**
```python
nir.LIF(
    tau=np.array(0.001),        # Single scalar value
    v_threshold=np.array(1.0),  # Single scalar value
    neuron_count=128,           # Explicit neuron count
)
```

## Root Cause

NIR's type inference relies on parameter array shapes to determine neuron counts:

```python
# From NIR source - infers neuron count from parameter shapes
def infer_neuron_count(self):
    return self.tau.shape[0] if self.tau.ndim > 0 else None
```

When parameters are scalars (0-d arrays), NIR cannot infer the neuron count and sets empty type annotations:
- `input_type: {'input': array([], dtype=float64)}`  (wrong)
- `output_type: {'output': array([], dtype=float64)}`  (wrong)

This causes type checking to fail with errors like:
```
ValueError: Type inference error: type mismatch: fc1.output: [[128]] -> lif.input: []
```

### 1. **Memory Inefficiency**
- A scalar parameter becomes an array of N identical values
- For large layers (e.g., 1024 neurons), this creates 1024x memory overhead
- Particularly problematic for neuromorphic hardware with limited memory

### 2. **Hardware Deployment Issues**
- Neuromorphic chips are optimized for shared scalar parameters
- Broadcasting forces hardware to store/access redundant identical weights
- Reduces effective model capacity on memory-constrained devices

### 3. **Export Complexity**
- Forces export tools to choose between efficiency and type safety
- Current workarounds require parameter broadcasting that loses semantic meaning

### Option 1: Explicit Neuron Count Parameter
Add an optional `neuron_count` parameter to neuron nodes:

```python
nir.LIF(
    tau=np.array(0.001),     # Scalar - applies to all neurons
    v_threshold=np.array(1.0), 
    neuron_count=128,        # Explicit count for type inference
    input_type={'input': np.array([128])},   # Auto-generated from neuron_count
    output_type={'output': np.array([128])}, # Auto-generated from neuron_count
)
```

### Option 2: Broadcast Semantics
Add support for broadcast-compatible parameters:

```python
nir.LIF(
    tau=np.array(0.001),           # Scalar broadcasts to all neurons
    v_threshold=np.array([1.0, 1.2, 1.0]),  # Per-neuron values where needed
    broadcast_params=['tau'],       # Explicit broadcast annotation
)
```

### Option 3: Shape Inference Context
Allow type inference to use context from connected layers:

```python
# Infer neuron count from connected Affine layer output shape
fc1 = nir.Affine(weight=np.random.randn(128, 784))  # 128 outputs
lif = nir.LIF(tau=np.array(0.001))  # Infer 128 neurons from fc1 connection
```

## Current Workaround

Export tools currently broadcast scalars to satisfy type checking:

```python
def to_array(val, n_neurons):
    """Broadcast scalar to array of n_neurons length"""
    if np.isscalar(val):
        return np.full(n_neurons, float(val))
    # ... handle existing arrays
```

This works but loses the semantic meaning and efficiency of scalar parameters.

## Test Case

```python
import nir
import numpy as np

# This should work without broadcasting
lif_scalar = nir.LIF(
    tau=np.array(0.001),        # Scalar parameter
    v_threshold=np.array(1.0),
    neuron_count=128,           # Proposed: explicit neuron count
)

# Type inference should succeed
graph = nir.NIRGraph(
    nodes={'lif': lif_scalar, ...},
    edges=[...]
)
# Should pass: nir.read(file, type_check=True)
```

## References

- **snnTorch export issue**: Scalar parameters must be broadcast to arrays for NIR compatibility
- **Hardware efficiency**: Neuromorphic chips (Intel Loihi, SpiNNaker) optimize for shared scalar parameters
- **Memory impact**: Large models with broadcast parameters can exceed hardware memory limits

## Environment

- NIR version: 1.0.7
- nirtorch version: 2.0.5
- Use case: snnTorch → NIR → neuromorphic hardware deployment

---

**Expected Outcome**: NIR should support efficient scalar parameters while maintaining type safety, enabling optimal hardware deployment without unnecessary memory overhead.

siddiquifaras:
Scalar neuron parameters (0-d arrays) produce empty type annotations that fail type checking. When infer_types() detects a type mismatch where the post node has scalar-derived types (size == 0), it now adopts the predecessor's output type instead of raising an error. Parameters remain untouched -- only type annotations are resolved.

Also guards serialization against h5py's inability to gzip-compress scalar (0-d) datasets.

Closes #176

### DIFF
--- a/nir/ir/graph.py
+++ b/nir/ir/graph.py
@@ -324,8 +324,7 @@ class NIRGraph(NIRNode):
                 # by adopting the predecessor's output type. Parameters are untouched.
                 post_values = list(post_node.input_type.values())
                 is_scalar_type = all(
-                    isinstance(v, np.ndarray) and v.size == 0
-                    for v in post_values
+                    isinstance(v, np.ndarray) and v.size == 0 for v in post_values
                 )
                 if is_scalar_type:
                     post_node.input_type = {
--- a/nir/ir/graph.py
+++ b/nir/ir/graph.py
@@ -318,16 +318,37 @@ class NIRGraph(NIRNode):
                     for k, v in pre_node.output_type.items()
                 }
             elif type_mismatch:
-                # set post input_type to be the same as pre output_type
-                pre_repr = (
-                    f"{pre_key}.output: {np.array(list(pre_node.output_type.values()))}"
+                # Check if post node has scalar-derived types (empty shape arrays).
+                # Scalar parameters (0-d arrays) produce input_type/output_type with
+                # empty arrays via np.array(param.shape) where shape=(). Resolve these
+                # by adopting the predecessor's output type. Parameters are untouched.
+                post_values = list(post_node.input_type.values())
+                is_scalar_type = all(
+                    isinstance(v, np.ndarray) and v.size == 0
+                    for v in post_values
                 )
-                post_repr = (
-                    f"{post_key}.input: {np.array(list(post_node.input_type.values()))}"
-                )
-                raise ValueError(
-                    f"Type inference error: type mismatch: {pre_repr} -> {post_repr}"
-                )
+                if is_scalar_type:
+                    post_node.input_type = {
+                        k.replace("output", "input"): v.copy()
+                        for k, v in pre_node.output_type.items()
+                    }
+                    post_node.output_type = {
+                        k.replace("input", "output"): v.copy()
+                        for k, v in post_node.input_type.items()
+                    }
+                else:
+                    pre_repr = (
+                        f"{pre_key}.output: "
+                        f"{np.array(list(pre_node.output_type.values()))}"
+                    )
+                    post_repr = (
+                        f"{post_key}.input: "
+                        f"{np.array(list(post_node.input_type.values()))}"
+                    )
+                    raise ValueError(
+                        f"Type inference error: type mismatch: "
+                        f"{pre_repr} -> {post_repr}"
+                    )
 
             # make sure that output nodes have output_type = input_type
             if isinstance(post_node, Output):
--- a/nir/serialization.py
+++ b/nir/serialization.py
@@ -242,13 +242,16 @@ def write(
             elif isinstance(v, str):
                 group.create_dataset(k, data=v, dtype=h5py.string_dtype())
             elif isinstance(v, np.ndarray):
-                group.create_dataset(
-                    k,
-                    data=v,
-                    dtype=v.dtype,
-                    compression=compression,
-                    compression_opts=compression_opts,
-                )
+                if v.ndim > 0:
+                    group.create_dataset(
+                        k,
+                        data=v,
+                        dtype=v.dtype,
+                        compression=compression,
+                        compression_opts=compression_opts,
+                    )
+                else:
+                    group.create_dataset(k, data=v, dtype=v.dtype)
             elif isinstance(v, dict):
                 write_recursive(group.create_group(str(k)), v)
             else:
--- a/tests/test_ir.py
+++ b/tests/test_ir.py
@@ -755,6 +755,206 @@ def test_type_check_recurrent():
         )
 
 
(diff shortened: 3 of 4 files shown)


nir/ir/graph.py
```diff
@@ -318,16 +318,37 @@ def infer_types(self):
...
+                # empty arrays via np.array(param.shape) where shape=(). Resolve these
+                # by adopting the predecessor's output type. Parameters are untouched.
+                post_values = list(post_node.input_type.values())
+                is_scalar_type = all(
+                    isinstance(v, np.ndarray) and v.size == 0
+                    for v in post_values
                 )
+                if is_scalar_type:
+                    post_node.input_type = {
+                        k.replace("output", "input"): v.copy()
```
Jegp, line 373: Let me make sure I'm understanding this: you're taking the dict `pre_node.output.items()`, and then replacing the "output" node with "input" copying `v` as the value. Is this a good strategy? I get the idea that we want to copy the output parameter, but, in principle, one node can have multiple outputs.

I'm not sure I have a great solution for this, but it's worth thinking a bit about. Would it be an option to build a new dictionary that only aligns the input? It's a bit of a headache because the ports are slightly underspecified (i. e. people can (ab)use them in whichever way), but I'd like to avoid constraining that too much if it's not necessary.

Jegp (changes requested):
Excellent addition, thanks.

I'd like to hear your input on the issue with copying over the output dictionary for input ports. It would also be helpful to have a test where we have nodes with multiple outputs/inputs, to make sure that part is well behaved.

nir/ir/graph.py
siddiquifaras: Thanks for the feedback. The `k.replace("output", "input")` pattern is consistent with the existing implementation in `infer_types()` (lines 316–318), which uses the same approach for undefined input types.


### 1. Current NIR implementation

All nodes use single-port naming (`{"input": ...}`, `{"output": ...}`), and multi port nodes currently raise `NotImplementedError` in `check_types()` (lines 216–218). In this context, the string replacement works reliably.

### 2. Future multi port support

When multi port nodes are introduced, NIR edges connect entire nodes (all ports to all ports), not individual ports. The type check at line 307 enforces port count consistency:

`len(post_node.input_type) != len(pre_node.output_type)`

Copying the full `output_type` dict to `input_type` therefore remains correct.

### 3. Port naming convention

The approach also extends to indexed ports (e.g., `output_0 → input_0`, `output_1 → input_1`) as long as the naming convention retains the `"output"` / `"input"` substrings.

i've added a test (`test_scalar_lif_port_naming`) to confirm that port naming is preserved and to document this behavior for future multi port support.

PR #191, opened 2026-06-30
Migrated type and metadata field to NIRNode dataclass field factory

Jegp:
Uses a nifty Python 3.10 dataclass feature to add default dict constructors to the `input_type`, `output_type`, and `metadata` fields in `NIRNode`, which removes a lot of redundant lines in inherited classes

### DIFF
--- a/nir/ir/graph.py
+++ b/nir/ir/graph.py
@@ -1,5 +1,5 @@
 from collections import Counter
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import Any, Dict, Optional
 
 import numpy as np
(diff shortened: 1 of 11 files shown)


nir/ir/graph.py
```diff
@@ -557,6 +552,8 @@ class Identity(NIRNode):
     This is a virtual node, which allows for the identity operation.
     """
 
+    # Shape of incoming data. Overrides the keyword-only input_type from
+    # NIRNode so it can be passed positionally
     input_type: Types
```
benkroehs, line 554: this needs to be dropped, no?

benkroehs (commented):
looks good, just one minor question

nir/ir/graph.py
Jegp: Not this one: Identity still needs input_type as a constructor argument. Unlike the base-class fields, Identity is defined by its input shape: it takes input_type positionally and `__post_init__` sets `output_type = input_type`. But I thought about it and I can see that we could drop some dead type params in `NIRGraph.__init__`. I'll push that and proceed. Thanks for the comment :)

PR #192, opened 2026-07-02
chore: Syntactic sugar for NIRData

benkroehs:
With this PR I want to introduce some syntactic sugar to simplify the access of die different observables and nodes in the `NIRGraphData`. Concretely this means that something like
`spikes = nir_graph_data.nodes['lif'].observables['spikes']`
can now be written as
`spikes = nir_graph_data['lif']['spikes']`

What do you think?

### DIFF
--- a/nir/data_ir/graph.py
+++ b/nir/data_ir/graph.py
@@ -30,6 +30,12 @@ class TimeGriddedData:
                 "and of type np.ndarray"
             )
 
+    def __getitem__(self, idx):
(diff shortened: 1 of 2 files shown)


Jegp (approved):
Looks good! My only comment would be to add a test for this, but the change is relatively harmless so feel free to go ahead.

benkroehs:
I added a small test that should cover the introduced getter and setter functionality.

PR #194, opened 2026-08-14
feat: Generalize NIRData API and Conversions

benkroehs:
With this PR I propose adding the following features to NIRData:
* Frameworks do not need to check in their from-NIRData version for the type of the observable data (`TimeGriddedData` or `(Valued)EventData`) , they can just call `data.get_event()`
* The continuous `time_shift` parameter is substituted by a boolean `dynamic_before_transition` to represent the two existing implementations for time-gridded simulators: propagating the the dynamics first and checking if the threshold has been exceeded after or on the contrary first checking for the threshold-crossing and then calculating the dynamics accordingly
* The dimension order for `TimeGriddedData` is no longer fixed but stored in the `dimension_order` attribute. The conversion happens lazily if a different format is requested. The `__getitem__` and `__setitem__` functions use per default the most popular ordering (time, batch, neuron), but this is flexible.

An exemplary implementation of these changes can be seen in the [most recent NIRTorch PR](https://github.com/neuromorphs/NIRTorch/pull/53).

### DIFF
--- a/nir/data_ir/graph.py
+++ b/nir/data_ir/graph.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 from dataclasses import dataclass
 from typing import Dict, Union
+import warnings
 import numpy as np
 from nir.ir import NIRGraph, NIRNode
 
@@ -18,16 +19,21 @@ class TimeGriddedData:
         Input data. For binary data the dtype should be bool.
     dt: float
         Time step size.
+    dynamic_before_transition: bool, optional
+        If True, it is assumed that the framework evolves the state (e.g.,
+        membrane potential) of the neurons before checking if the threshold has
+        been crossed and generating an event (transition). If False, the state
+        is updated after the event generation.
     """
 
     data: np.ndarray
     dt: float  # pylint: disable=invalid-name
+    dynamic_before_transition: bool = True
 
     def __post_init__(self):
         if not isinstance(self.data, np.ndarray) or self.data.ndim != 3:
             raise ValueError(
-                "Data must be of shape (n_samples, n_time_steps, n_neurons)"
-                "and of type np.ndarray"
+                "Data must be of shape (n_samples, n_time_steps, n_neurons)and of type np.ndarray"
             )
 
     def __getitem__(self, idx):
@@ -56,15 +62,12 @@ class TimeGriddedData:
     def t_max(self):
         return self.n_time_steps * self.dt
 
-    def to_event(self, n_events: int, time_shift: float = 0.0) -> EventData:
+    def to_event(self, n_events: int) -> EventData:
         """
         Arguments
         ---------
         n_spikes : int
             Maximum number of events stored for each neuron.
-        time_shift : float, optional
-            Shift the event times by this value from the beginning of each time
-            step. Must be in interval [0, dt). Default is 0.0.
         """
 
         if not self.data.dtype == bool:
@@ -72,8 +75,6 @@ class TimeGriddedData:
         idx = np.full((self.n_samples, n_events), -1)
         time = np.full((self.n_samples, n_events), np.inf)
 
-        if time_shift < 0 or time_shift >= self.dt:
-            raise ValueError("time_shift must be in interval [0, dt)")
         for sample in range(self.n_samples):
             time_step, neuron = np.where(self.data[sample])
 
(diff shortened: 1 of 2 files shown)


nir/data_ir/graph.py
```diff
@@ -124,27 +230,76 @@ def shape(self):
     def n_samples(self):
         return self.idx.shape[0]
 
-    def to_time_gridded(
-        self, dt: float  # pylint: disable=invalid-name
+    def get_event(self, n_events: int | None) -> EventData:
```
Jegp, line 233: Is it correct that `EventData` is collecting a series (what you call sample) of time-stepped events? So, that's basically a "batched" coordinate AER view, correct?

And is it correct that `n_events` cap each sample to `n_events`? If so, 
1. why not ask for a lower cap as well? Would you want to "slice" the samples, like you'd slice a normal list?
2. and if you are more or less slicing, what's the logic below good for? Why not just do `idx[:, start:end]`? I can see you're doing checks for invalid events, but I'm puzzled why that has to happen in the getter. Shouldn't that be a constructor check to make sure the data itself is well-formed upon construction?

nir/data_ir/graph.py
```diff
@@ -173,9 +328,44 @@ def __post_init__(self):
         if self.idx.shape != self.time.shape or self.idx.shape != self.value.shape:
             raise ValueError("idx, time and value must have the same shape")
 
-    def to_time_gridded(
+    def get_event(self, n_events: int | None) -> ValuedEventData:
```
Jegp, line 331: Same comment as for L233

nir/data_ir/graph.py
Jegp, line 419: Ruff complains :P

Jegp (changes requested):
Very very nice. I have one question around the use of coordinate-based representations in the `EventData` class. And there's a merge conflict after the Ruff branch merged.

Jegp:
Additionally, I wonder whether it would make sense to actually model the relationship between the NIRNode and the NIRData. Currently, if I'm understanding it correctly, [you're simply checking that the dimensionality of the data fits with the shape of the node](https://github.com/neuromorphs/NIR/blob/main/nir/data_ir/graph.py#L236). Wouldn't it be nicer to have a direct link, so you can access the node in the data format? Something like an augmented `NIRNode` where you add a `data` entry, or similar? Like

```python

@dataclass
class NIRNodeData(NIRNode):
    observables: dict[str, EventData | TimeGriddedData]
    
    def check_observables(self):
        """
        Check that the shapes of the observables match the node's output shapes
        """
        output_shape = self.output_type["output"]
        return all(obs.n_neurons == output_shape for obs in self.observables.values())
```

Wouldn't that make the relationship unambiguous?

muffgaga:
(Ben and I are just sitting together and thinking about this :) .)

Regarding your last comment @Jegp — you suggest `NIRNodeData` to inherit from `NIRNode` and construct with an `NIRNode` instance reference to have an explicit link to a existing node in the `NIRGraph`?
This does not look correct to me — primitives are `NIRNode`s (…and `NIRGraph`, nesting is possible which is nice :D); however, the existing `NIRNodeData` doesn't "process" anything, it's not a vertex in the signal-flow graph.
If you suggest the inverse, having "data" in the `NIRNode`s → yes, that does not sound wrong — however, in this case I don't think it should be inheritance ("is-a") but rather composition ("has-a").
In that case we could have a ternary `.data` member that provides `None | ObservableData` — `None` would provide the sentinel value for "wasn't recorded/observed" and 0-length `EventData` would mean "no spike" observed.

Advantage: no separate observable data structure that has a "string-based" link to the topology data structure (and no possibility to have a mismatch in "keys", i.e. no need for a "key-match" check).
Disadvantage: one more member variable in the previously-only-topology data structure and the NIRGraph would mix topology and observables; `from/to_nir_data()` would be weird and NIR implementations for individual backends have to provide the variable (so topology im/export-only backends might be more "confusing" to the user as you would get None by default (which might be ok)). 

We don't have a strong opinion ;) …

nir/data_ir/graph.py
benkroehs: So EventData is not time-stepped data but continuous/event-based. For each spike you have a `time` when it occured and and index of the spiking neuron `idx`.
Maybe the phrase "each event is discrete" in the docstring is misleading. What I meant by that is that an event does not has a length but happens at one point in time.

And yes, `n_events` cap the events per sample to this number. One could say that `n_events` is the number of events per sample. If a sample has 5 spikes but `n_events is 10, the last 5 events would be "empty/invalid events" with `idx=-1`. But if there were 15 spikes, the last 5 would be dropped. I implemented this for a better data handling.

Does this clear things up on your side?

PR #196, opened 2026-09-15
Add spikeforge to the framework support table

w4ffl35:
Adds [spikeforge](https://github.com/Capsize-Games/spikeforge) (BSD-3-Clause, `pip install spikeforge`) to the framework support table. It's an SNN toolkit built on snnTorch/PyTorch that reads and writes NIR through `spikeforge.nir_bridge`.

**One thing I'd like your judgement on, rather than quietly ticking the box.**

I've marked *Read from NIR* as ✓, but spikeforge's read path differs from the other rows and you may prefer ⬚. The other frameworks import a NIR graph into their own native model. spikeforge loads the graph and **executes it directly** through an independent interpreter — it does not reconstruct a native module:

```python
from spikeforge.nir_bridge.ingest import load_external, interpret_file

graph = load_external("model.nir")       # NIRGraph, from any framework
result = interpret_file("model.nir", spikes)   # executed, never touching snnTorch
```

So a NIR file written by Norse or snnTorch runs in spikeforge and produces outputs, which I think is what the column means to a user — but it is genuinely not "convert to native". Happy to change it to ⬚ if you read the column the other way. Any node kind outside the shared vocabulary raises a typed error naming the node rather than being skipped.

**Write** is the conventional direction: `to_nir(spec, module)` emits a graph, verified against the 20-node `conv_net` preset.

**Why the row might be worth having beyond one more entry:** spikeforge re-executes every exported graph through that same independent interpreter and reports per-tensor numerical drift against the source model, so an export can be *validated* rather than assumed. As far as I can tell none of the listed frameworks does this, and it's a NIR-specific capability rather than a general feature:

```bash
pip install "spikeforge[nir]"
spikeforge-verify validate --topology conv_net
```

That exits non-zero outside tolerance. On `conv_net` today the worst deviation is `1.4e-07` mean-abs on the `lif1` membrane.

**Examples column:** every other row points at a neuroir.org docs page. spikeforge has none, so I've linked the runnable example in-repo (`examples/04_nir_export_validate.py`, verified end to end on a clean install). If you'd like a proper docs page in the NIR docs tree instead, I'm glad to write one in a follow-up PR — I didn't want to add pages to your docs uninvited in a first contribution.

Docs: <https://docs.spikeforge.net/interpreter-spine>


Jegp (approved):
Thanks for the addition. I appreciate your question about the interpretation of reading/writing, but we're very loose here: if you can load a graph and run if, I'd tick the box, no matter how you do it. Same for writing. 

I see you're using snnTorch behind the scenes, but what I think sets spikeforge apart is that you *could* use another simulator one day, the API isn't strictly tied to a simulator.

I'm happy to approve and merge as-is. My only suggestion would be to add an example in the [example pages](https://neuroir.org/docs/examples/) so it's clear how your framework is used. Basically, you can just copy in the code you wrote in this PR :-)

